PR #1, opened 2019-11-12
Initial commit of rimage tool

dcpleung:
This is the initial commit to add the `rimage` tool. Zephyr is in the process of enabling a BSP for the APL DSP on UP Squared board, which requires rimage to build the firmware. Discussion with @lgirdwood resulted in this work. Note that the files under `src/include/sof/` are only needed when building without the SOF tree. If the SOF tree path is passed to `configure` with `--with-sof-source=<path>`, the same-name files in the SOF tree are used instead.

This utilizes autotool and thus the tool can be built with configure/make.


dbaluta:
@dcpleung is there possible to also import the git history from old repo?

dcpleung:
@dbaluta Let me try.

cujomalainey:
Only ways I can think of doing that is either copying the whole history and deleting everything else or manually cherry-picking the history of the rimage folder and handling backports where necessary

lgirdwood:
I dont think it's a big deal if history cannot be easily preserved here as it will still be in the main sof repo master branch.

dcpleung:
Updated it with history of rimage (files under SOF's rimage/).

cujomalainey:
I don't see the keys committed here, will they be passed in by the builders?

dbaluta:
@dcpleung thanks for taking care of this! Looks good to me.

dcpleung:
@cujomalainey The keys are still in the SOF repo, and will be passed as command line arguments.

cujomalainey:
@dcpleung thanks :) what is the motivation for splitting this out of the SOF repo? I thought the reason for merging the SOFT repo was so we had less issues with keeping ABIs in sync across tools, this seems like a step backward since there are copies of the IPC ABI in these commits.

dcpleung:
Zephyr is in the process of enabling a BSP for the APL DSP on UP Squared board, which requires rimage to build the firmware. Discussion with @lgirdwood resulted in this work. Note that the files under `src/include/sof/` are only needed when building without the SOF tree. If the SOF tree path is passed to `configure` with `--with-sof-source=<path>`, the same-name files in the SOF tree are used instead.

cujomalainey:
Sounds good to me, thanks for the information

plbossart:
> Zephyr is in the process of enabling a BSP for the APL DSP on UP Squared board, which requires rimage to build the firmware. Discussion with @lgirdwood resulted in this work. Note that the files under `src/include/sof/` are only needed when building without the SOF tree. If the SOF tree path is passed to `configure` with `--with-sof-source=<path>`, the same-name files in the SOF tree are used instead.

Please add this to the PR description...

my concern is that once you have rimage, you'll want to use actual code from the SOF tree, so what's the plan here? The last split and merge of SOFT was a nightmare, I'd like to make sure we have a direction that's well agreed on. Any solution that impacts our ability to bisect is a NAK.

jajanusz:
Why was it merged? I see there a bit of mess. Fe. 2 buildsystems - automake and  cmake, we can go with just cmake as both sof and zephyr use it.

plbossart:
> Why was it merged? I see there a bit of mess. Fe. 2 buildsystems - automake and cmake, we can go with just cmake as both sof and zephyr use it.

agree.

lgirdwood:
I'm guessing this is an issue for Windows ?

Userspace tooling is traditionally built with autotools since it has no extra dependencies and easy for the distros. I've no objections to cmake if someone wants to copy the current camke file over.

marc-hb:
Anyone looking for the rimage history and hitting the first, b537522 commit in this repo (a.k.a. the "end of the road") must continue to https://github.com/thesofproject/soft/commits/72d797d9ddae7 (and not to https://github.com/thesofproject/tools which is not a thing, thanks @ranj063 for the tip)

> Only ways I can think of doing that is either copying the whole history and deleting everything else or manually cherry-picking the history of the rimage folder and handling backports where necessary

I remember `git filter-branch` could do pretty much anything - including shooting in an unlimited number feet. Now I see `git help filter-branch` recommends `git filter-repo` instead, never tried the new one.

marc-hb:
> Anyone looking for the rimage history [...] must continue to https://github.com/thesofproject/soft/commits/72d797d9ddae7

Funny enough I just discovered that this sof-tools.git history is also in sof.git thanks to @cujomalainey (thanks!)

Complete story at
- https://github.com/thesofproject/sof/issues/8178#issuecomment-1710615354

... where merging rimage back is discussed.

PR #3, opened 2020-03-25
Use CMake to build rimage

dcpleung:
This adds the necessary bits to allow building rimage with CMake with similar configure options as the just removed autotools files.

Also update the rimage source from SOF master (@ 4e87899f25b0c859169c1dfd4dd9f9fd65d48ad2).


lgirdwood:
@jajanusz good for you ?

PR #6, opened 2020-04-07
Remove unused 'image' argument in elf_find_section()

ktrzcinx:
Unused arguments shouldn't be passed to function because it
makes code messy.

### DIFF
--- a/src/elf.c
+++ b/src/elf.c
@@ -76,7 +76,7 @@ static int elf_read_sections(struct image *image, struct module *module,
 		module->bss_end = 0;
 	} else {
 		/* find manifest module data */
-		module->bss_index = elf_find_section(image, module, ".bss");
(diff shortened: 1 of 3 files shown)


src/elf.c
```diff
@@ -464,11 +464,10 @@ int elf_validate_modules(struct image *image)
 	return 0;
 }
 
-int elf_find_section(struct image *image, struct module *module,
-		     const char *name)
+int elf_find_section(const struct module *module, const char *name)
 {
 	Elf32_Ehdr *hdr = &module->hdr;
-	Elf32_Shdr *section, *s;
+	const Elf32_Shdr *section, *s;
```
paulstelian97, line 470: This addition of "const" would normally warrant a separate commit just for itself. I won't impose this though.

PR #13, opened 2020-04-22
Ext mnf optional

ktrzcinx:
https://github.com/thesofproject/rimage/issues/12

Know logs for case when ext_mnf is not present loooks like:

```
warning: can't find section .fw_metadata in module /mnt/c/projects/sof_kt/build/sof-cnl
warning: unable to write extended manifest, -125
```

instead of:
```
warning: can't find section .fw_metadata in module /home/lrg/work/zephyrproject/zephyr/build/zephyr/zephyr.elf.mod
error: section .fw_metadata can't be found
error: failed to read .fw_metadata section content, code -22
warning: unable to write extended manifest, -22
```

For case with ext_man, there is no any difference in logs

### DIFF
--- a/src/rimage.c
+++ b/src/rimage.c
@@ -42,6 +42,7 @@ static void usage(char *name)
 	fprintf(stdout, "\t -x set xcc module offset\n");
 	fprintf(stdout, "\t -f firmware version = x.y\n");
 	fprintf(stdout, "\t -b build version\n");
+	fprintf(stdout, "\t -e build extended manifest\n");
 	exit(0);
 }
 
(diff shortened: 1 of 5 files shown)


src/rimage.c
```diff
@@ -193,7 +193,7 @@ int main(int argc, char *argv[])
 
 	ret = ext_man_write(&image);
 	if (ret < 0) {
-		fprintf(stderr, "warning: unable to write extended manifest, %d\n",
+		fprintf(stdout, "warning: unable to write extended manifest, %d\n",
```
lgirdwood, line 196: This should really only be printf if we ask rimage to generate the ext manifest. i.e. we have a rimage cmd line option for ext manifest generation. i.e.(using -x for ext manifest generation)
1) ```rimage -x``` -> emits error message and stops if ext man not found
2) ```rimage``` -> no error or warning if ext manifest not found.

ktrzcinx: `-e` flag is used, because `-x` is already in use

PR #16, opened 2020-05-13
headers: Move headers to rimage namespace

ktrzcinx:
Because there is possibility to reference headers from this repo
from another one, there is a need to put every header into rimage
namespace. Otherwise reference to headers from duplicated
namespace (like kernel) from external repo may be ambiguous.

This PR will need corresponding one from FW side after merge


jajanusz:
I agree that there should be no files in 'include/sof` folder coming from this repo.
However if these are no longier sof-specific symbols (cos they live in generic rimage directory), probably they should be prefixed with RIMAGE_* instead of SOF_*.

Are these symbols going to be used by all users of rimage (sof, zephyr, etc)? Or maybe these are just symbols specific for our flavour of rimage-composed FW image, then we should have something like `include/rimage/sof` folder.

@lgirdwood please provide some input

lgirdwood:
@ktrzcinx are you saying we have duplicated headers today between SOF and rimage ?

ktrzcinx:
@lgirdwood not exactly. When some header have the same file name in rimage and repo and in FW, then build system is ambiguous which one should be used. Such a situation is for example for ext_manifest.h, there are:
`*rimage/src/include/sof/*kernel/ext_manifest.h` and
`*sof/src/include/*kernel/ext_manifest.h`

`sof/src/include/` obviously is in FW build path, and it's ok, but `rimage/src/include/sof/` also is there, what leads to this hesitation. When I want to include header from rimage repo, then in FW source code I should write it explicity to solve this situation - `#include <rimage/...>`, what will be after changes from this PR. Moreover guard symbols in `#ifndef/#define/#endif` also must be unique between this file.

Second issue:
After splitting rimage repo, then there is some double for `SOF` prefixes in `kernel/fw.h` and  `user/manifest.h` (@jajanusz question). Should I replace them `SOF` to `RIMAGE` as more generic definitions, or leave as it is?

lgirdwood:
@ktrzcinx ok, so we have some files with conflicting file names and guards. Can you con firm we are not duplicating structures or other things ?

ktrzcinx:
@lgirdwood as I see only structs `ext_man_header` and `ext_man_elem` definitions are currently duplicated because of parallel work on rimage repo split and ext_manifest deployment. Changes from this PR are needed to remove this duplications.

lgirdwood:
@ktrzcinx ok, got you now. @jajanusz any comments from you ?

jajanusz:
It's step in the right direction

PR #17, opened 2020-05-13
ext_manifest: Add UUID dictionary

ktrzcinx:
Such a dictionary will be used in kernel code to translate UUID value
to UUID dictionary key used in DSP in runtime.
Translation will be used for the process component creation and
runtime trace filtering.
Because there is no way to put right size to this extended element
header, list of uuid entries are located in separate section to
allow convenient size calculation from rimage level.
Extract common header with simple, generic preprocessor functions.

This PR is related with https://github.com/thesofproject/sof/pull/2914.

### DIFF
--- a/src/ext_manifest.c
+++ b/src/ext_manifest.c
@@ -128,6 +128,8 @@ static int ext_man_build(const struct module *module,
 	*dst_buff = (struct ext_man_header *)buffer;
 
 out:
+	if (!sec_buffer)
+		free(sec_buffer);
 	return ret;
 }
 
--- a/src/ext_manifest.c
+++ b/src/ext_manifest.c
@@ -11,6 +11,7 @@
 
 #include "kernel/ext_manifest_gen.h"
 #include "kernel/ext_manifest.h"
+#include "common.h"
 #include "rimage.h"
 
 const struct ext_man_header ext_man_template = {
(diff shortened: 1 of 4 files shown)


src/ext_manifest.c
```diff
@@ -81,6 +82,87 @@ static int ext_man_validate(uint32_t section_size, const void *section_data)
...
+ * Return `section_name` size or error code.
+ */
+static int ext_man_update_element_with_list(const struct module *module,
+					    int32_t *metadata_size,
+					    uint8_t **metadata_buff,
+					    enum ext_man_elem_type elem_type,
+					    const char *section_name)
+{
+	struct ext_man_elem_header* head = NULL;
+	uint8_t *meta_buff = *metadata_buff;
```
lgirdwood, line 97: Can you confirm we are converting a UUID to a memory address (as stated in commit message) ? This is a security hole, we cannot pass in addresses from outside of FW, we can however pass in data we can validate like an index in an array. @lbetlej @mmaka1 any comments here ?

kv2019i: @lgirdwood The code is not merged yet to kernel, but my understanding what is passed is a key to the dictionary, not an actual address. @keyonjie @ktrzcinx can you clarify?

keyonjie: Yes, it is only a index-liked "key", not the real memory address, those real 16 Bytes UUIDs are not stored in DSP SRAM actually. @mmaka1 please correct me if I am wrong.

ktrzcinx: This is address of UUID entry, from invalid memory region (0x1FFFA000 is already used as start address, but it's fully configurable in linker script) - so dereference leads to firmware crash. This address is used as UUID element identifier/representation from FW side because full UUID value is inaccessible from there. So this address is UUID dictionary key value.

lgirdwood: ok, can you update the commit messages, inline comments and PR description to use key/index rather than address.

ktrzcinx: done

ktrzcinx:
This solution was rejected, full UUID will be passed during building topology.

PR #21, opened 2020-09-09
Move platform configuration to config files

ktrzcinx:
Move platform configuration to separate file. New platform addition shouldn't be related with rimage recomplilation. Moreover new way of configuration allow easily add default value to any field.

There are a few possible config file formats available to choose:

ini - problem with nested structures
json - lack of comments and poor diff tool output
yaml - advanced language, complicated parsers
toml - supported comments, any problem with nested structures, nice diff tool output

Build test results: https://github.com/thesofproject/sof/pull/3425


src/adsp_config.c
```diff
@@ -0,0 +1,1158 @@
...
+
+	/* check "version" key */
+	arr = toml_array_in(toml, "version");
+	if (!arr || toml_array_type(arr) != 'i'  || toml_array_nelem(arr) != 2 ||
+	    toml_array_kind(arr) != 'v') {
+		ret = log_err(-EINVAL, "error: toml parse key 'version' failed\n");
+		goto error;
+	}
+
+	for(i = 0; i < ARRAY_SIZE(css_version); ++i){
```
lgirdwood, line 1107: Missing space after for(

src/adsp_config.c
```diff
@@ -0,0 +1,1158 @@
+/* SPDX-License-Identifier: BSD-3-Clause
```
lgirdwood, line 1: This file needs a lot more inline comments within each function explaining what is going on - difficult to review otherwise.

lgirdwood (changes requested):
I think the TOML format looks good and is easy to follow.
@jajanusz any comments here on how cmake deal with git submodules ? i.e. SOF has rimage submodule and now rimage has TOML submodule. I guess cmake needs to know how to take care of this ?

src/adsp_config.c
ktrzcinx: Added much more comments :)

src/adsp_config.c
ktrzcinx: Added, btw shouldn't we add checkpach to PR in this repo ?

ktrzcinx:
@lgirdwood Should I redirect git-submodule with toml parser to thesofproject mirror of tomlc99 repo ?

src/adsp_config.c
lgirdwood: Sounds like a good idea

lgirdwood:
> @lgirdwood Should I redirect git-submodule with toml parser to thesofproject mirror of tomlc99 repo ?

No we should use the default toml repo. Do we have any patches on top of TOML  ?

ktrzcinx:
>Do we have any patches on top of TOML ?

No, we don't.

lgirdwood:
Ok, use the official toml repo. We can always tag a version if needed.

lgirdwood:
@ktrzcinx do you have a ETA of when we can remove the DNM tag ?

ktrzcinx:
There's no manifest v1.5 support yet, but we can merge it separately, this PR is big enough :)

lgirdwood:
ok, so this wont break any existing cavs1.5 platforms ?

ktrzcinx:
Old way of defining platform is not deleted, so it shouldn't break anything

lgirdwood (approved):

marc-hb:
Proposal to drop submodules and replace with "west": https://github.com/orgs/thesofproject/teams/sof-developers/discussions/46

PR #25, opened 2020-09-22
travis: Add checkpatch run

ktrzcinx:
This tool checks mainly spelling and code formatting.

Signed-off-by: Karol Trzcinski <karolx.trzcinski@linux.intel.com>


.travis.yml
```diff
@@ -12,4 +12,8 @@ jobs:
       before_install: docker pull thesofproject/sof && docker tag thesofproject/sof sof
       script:
         - echo -e '#!/bin/bash\nmkdir build && cd build && cmake .. && make' > build.sh && chmod +x build.sh
-        - docker run -i -t -v `pwd`:/home/sof/work/sof.git --user `id -u` sof ./build.sh
\ No newline at end of file
+        - docker run -i -t -v `pwd`:/home/sof/work/sof.git --user `id -u` sof ./build.sh
```
marc-hb, line 15: Use $(...) notation instead of legacy backticked `...`

https://github.com/koalaman/shellcheck/wiki/SC2006

.travis.yml
```diff
@@ -12,4 +12,8 @@ jobs:
...
+    - name: checkpatch
+      script:
+        - printf '%s, %s\n' "${TRAVIS_BRANCH}" "${TRAVIS_COMMIT_RANGE/.../..}"
+        - (set x; scripts/checkpatch.pl --no-tree --strict --codespell --no-signoff -g ${TRAVIS_COMMIT_RANGE/.../..})
```
marc-hb, line 19: I guess you copied the `set -x` trick from the one I added in `linux/.travis.yml`. The purpose of this `set -x` trick is to show the expanded value of `${TRAVIS_COMMIT_RANGE/.../..}`, which makes the `printf` unnecessary.

If you think showing the value of `"${TRAVIS_BRANCH}"` is useful too, then you might as well make it really useful and run a short `git log` instead, see example at: https://travis-ci.org/github/thesofproject/linux/jobs/728647259

marc-hb (changes requested):
I don't understand what Travis did in https://travis-ci.org/github/thesofproject/rimage/jobs/729278090

- The `set -x` trick should have shown the checkpatch command with its expanded argument but it did not? See a working `set -x` example in a different repo: https://travis-ci.org/github/thesofproject/linux/jobs/728647259

- checkpatch printed the usual `dictionary.txt` not found error (because Ubuntu 20 changed its location), then it prints 1500 spelling errors?! Any idea why?

BTW you can copy the "fix" for the `dictionary.txt` not found error from https://github.com/thesofproject/linux/pull/2455/files

.travis.yml
ktrzcinx: Updated

.travis.yml
ktrzcinx: It wasn't target of this PR, but as it's quite small change I add it here.

lgirdwood (approved):
@marc-hb all good now ?

marc-hb:
> checkpatch printed the usual dictionary.txt not found error (because Ubuntu 20 changed its location), 

That was wrong sorry, this is not why codespell failed. codespell is still failing.

If you look at "Build system information" in https://travis-ci.org/github/thesofproject/rimage/jobs/729537424 you can see Travis uses its Ubuntu 16.04 default. The symbolic link hack is not needed for 16.04. What's missing is the codespell package itself:

```
before_install:
  - sudo apt-get -y install codespell
```

PS: there's still one pair of backticks left but you're right it's was not part of this change, I got confused because it was so close to it.

lgirdwood:
@ktrzcinx lets tag v1.6 after this is merged. I can do this tomorrow and then we can pull into SOF v1.6 release.

marc-hb:
@lgirdwood this is a pure `.travis.yml` and checkpatch change.  I wonder why a release would care?

lgirdwood:
@marc-hb I want the CI to be consistent for rimage stable fixes too. i.e. any rimage stable-v1.6 fixes must still pass CI just like on master

marc-hb:
Fair enough but this patch should be trivial to cherry-pick to the release branch, no validation required. I'm not on top of the release timeline and urgency so don't pay attention to me, just my 2 cents.

lgirdwood:
@marc-hb we have the time. Waiting on @keyonjie and @aiChaoSONG confirming tomorrow morning that we are good to tag.

ktrzcinx:
> I don't understand what Travis did in https://travis-ci.org/github/thesofproject/rimage/jobs/729278090

Showed codespell errors in project dictionary (uploaded in this PR).

> there's still one pair of backticks left

Fixed


Checkpatch shows codespell error in project dictionary and in checkpatch.pl, there's no more any warnings about checkpatch setup (after project dictionary upload), so I guess it should be ok.

lgirdwood (approved):
It's failing on itself, but that's expected.

PR #26, opened 2020-09-22
Add more warnings to make command, treat warnings as errors

ktrzcinx:
CI build system should be more restrictive.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,5 @@ jobs:
     - name: "Build Test"
       before_install: docker pull thesofproject/sof && docker tag thesofproject/sof sof
       script:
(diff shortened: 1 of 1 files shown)


lgirdwood (approved):

marc-hb:
I think having`-Werr` only in CI makes sense because it can be too annoying locally for work in progress.

On the other hand, why not have `-Wall -Wmissing-prototypes -Wimplicit-fallthrough=3 -Wpointer-arith` in CMakeLists.txt for everyone so they don't get surprised at the last minute by CI results? Or slowed down because the a missing warning doesn't show them a bug they have. They're going to have to fix these warnings eventually anyway so the sooner they get them the better, no?

PR #27, opened 2020-09-23
Fix a couple crashes

marc-hb:
See commit messages.

### DIFF
--- a/src/rimage.c
+++ b/src/rimage.c
@@ -36,7 +36,7 @@ int main(int argc, char *argv[])
 	struct image image;
 	struct adsp *heap_adsp;
 	const char *adsp_config = NULL;
-	int opt, ret, i, elf_argc = 0;
+	int opt, ret, i, first_non_opt;
 	int imr_type = MAN_DEFAULT_IMR_TYPE;
 	int use_ext_man = 0;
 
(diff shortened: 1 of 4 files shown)


src/rimage.c
```diff
@@ -86,11 +86,12 @@ int main(int argc, char *argv[])
 			usage(argv[0]);
 			break;
 		default:
+		 /* getopt's default error message is good enough */
 			break;
 		}
 	}
 
-	elf_argc = optind;
+	const int first_non_opt = optind;
```
ktrzcinx, line 94: Everywhere in project we declare new variables at the begin of function, I guess we shouldn't make here exception and keep code style consistent. Of course, then const must be removed, but it's not so critical for local variables.

lgirdwood (commented):
@marc-hb thanks for these fixes, we can apply once code style issue is resolved. I will probably tag this as v1.6.

src/rimage.c
marc-hb: This is a not just a code style question, this is an important new feature added by C99 to reduce the chance of memory corruption. Uninitialized variables is a cause of crashes and security issues.

I understand xt-xcc doesn't support C99 so it probably requires declarations at the start of a block (not function) but do we really need to keep user space tools stuck in the past too? C99 is 20 years old now.

lgirdwood: @marc-hb can you update this now, we can discuss this after v1.6. Thanks

marc-hb: Done.

marc-hb:
submodule update submitted at https://github.com/thesofproject/sof/pull/3457

PR #28, opened 2020-10-15
manifest.c: assert buffer overflow in man_copy_sram()

marc-hb:
Does not solve https://github.com/thesofproject/sof/issues/3423 but
converts a buffer overflow and crash into a more usable assert.

Buffer overflow found by running rimage with valgrind and the binaries
provided by @singalsu

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -167,6 +167,8 @@ static int man_copy_sram(struct image *image, Elf32_Shdr *section,
 	void *buffer = image->fw_image + offset;
 	size_t count;
 
+	assert((uint64_t)offset + section->size <= image->adsp->image_size);
+
 	switch (section->type) {
 	case SHT_INIT_ARRAY:
 		/* fall through */


ktrzcinx (approved):

marc-hb:
Thanks @ktrzcinx for the prompt review.

Very short review times that leave no chance for everyone to comment are bad. In this case however I'm going to fast-track this one because:
- it helps with real, reported and reproduced crash (https://github.com/thesofproject/sof/issues/3423)
- it's not going to affect SOF anyway until the submodule is updated!
- it's a dead-simple, one-line change that is very easy to revert in the worst of cases
- @lgirdwood is on vacation

marc-hb:
@lgirdwood or some other admin can you please turn on "Allow rebase merging" in the settings of this repo? See screenshot:
https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuring-commit-rebasing-for-pull-requests

This PR created an "empty merge" which creates noise in the git log. This noise is especially pointless in a low traffic repo like this one and even more annoying considering this is a submodule.

lgirdwood:
@marc-hb rebase merging is allowed (and is now the only option for this repo), so could have been GH preventing non admins from using it ?

marc-hb:
>  so could have been GH preventing non admins from using it ?

Mmmm... I would be very surprised if some people can merge one way and others some other way.

Let's just keep an eye on it.

PR #29, opened 2020-10-15
pkcs1_5.c: don't pretend everything is fine when failing to sign

marc-hb:
v2: as requested by @ktrzcinx , I also cleaned up the code a bit. Now there are 2 commits:
- 1st commit changes the entire source of these functions but does not change the rimage compiled binary
- 2nd commit is the actual fix with just a one-line change in each function

v1:
Found by chance and reproduced by making a typo in the key's filename.

Note there is already an (unchanged) error message:
```
 pkcs: signing with key sof/keys/otc_private_key_3k.pemX
 error: can't open file sof/keys/otc_private_key_3k.pemX -2
 Firmware manifest and signing completed !
```
But now rimage does not pretend success and the build does not keep
going with an unsigned image like nothing happened.

### DIFF
--- a/src/pkcs1_5.c
+++ b/src/pkcs1_5.c
@@ -244,9 +244,10 @@ int ri_manifest_sign_v1_5(struct image *image)
 {
 	struct fw_image_manifest_v1_5 *man = image->fw_image;
 
-	pkcs_v1_5_sign_man_v1_5(image, man,
-				(void *)man + MAN_CSS_MAN_SIZE_V1_5,
-				image->image_end - sizeof(*man));
+	char *const data1 = (char *)man + MAN_CSS_MAN_SIZE_V1_5;
+	unsigned const size1 = image->image_end - sizeof(*man);
+
+	pkcs_v1_5_sign_man_v1_5(image, man, data1, size1);
 	return 0;
 }
 
@@ -254,13 +255,16 @@ int ri_manifest_sign_v1_8(struct image *image)
 {
 	struct fw_image_manifest_v1_8 *man = image->fw_image;
 
-	pkcs_v1_5_sign_man_v1_8(image, man, (void *)man + MAN_CSS_HDR_OFFSET,
-				sizeof(struct css_header_v1_8) -
-				(MAN_RSA_KEY_MODULUS_LEN +
-				 MAN_RSA_KEY_EXPONENT_LEN +
-				 MAN_RSA_SIGNATURE_LEN),
-				(void *)man + MAN_SIG_PKG_OFFSET_V1_8,
-				(man->css.size - man->css.header_len)
-				 * sizeof(uint32_t));
+	char *const data1 = (char *)man + MAN_CSS_HDR_OFFSET;
+	unsigned const size1 =
+		sizeof(struct css_header_v1_8) -
+		(MAN_RSA_KEY_MODULUS_LEN + MAN_RSA_KEY_EXPONENT_LEN +
+		 MAN_RSA_SIGNATURE_LEN);
+
+	char *const data2 = (char *)man + MAN_SIG_PKG_OFFSET_V1_8;
+	unsigned const size2 =
+		(man->css.size - man->css.header_len) * sizeof(uint32_t);
+
+	pkcs_v1_5_sign_man_v1_8(image, man, data1, size1, data2, size2);
 	return 0;
 }
--- a/src/pkcs1_5.c
+++ b/src/pkcs1_5.c
@@ -247,8 +247,7 @@ int ri_manifest_sign_v1_5(struct image *image)
 	char *const data1 = (char *)man + MAN_CSS_MAN_SIZE_V1_5;
 	unsigned const size1 = image->image_end - sizeof(*man);
 
-	pkcs_v1_5_sign_man_v1_5(image, man, data1, size1);
-	return 0;
+	return pkcs_v1_5_sign_man_v1_5(image, man, data1, size1);
 }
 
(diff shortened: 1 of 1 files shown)


src/pkcs1_5.c
```diff
@@ -244,16 +244,17 @@ int ri_manifest_sign_v1_5(struct image *image)
...
 }
 
 int ri_manifest_sign_v1_8(struct image *image)
 {
 	struct fw_image_manifest_v1_8 *man = image->fw_image;
 
+	return
 	pkcs_v1_5_sign_man_v1_8(image, man, (void *)man + MAN_CSS_HDR_OFFSET,
 				sizeof(struct css_header_v1_8) -
 				(MAN_RSA_KEY_MODULUS_LEN +
```
ktrzcinx, line 260: How nicely argument list :) Is guess you added this return statement in the line before to keep this following monster as it is, but it looks quite odd (this function name is not so long, especially for 100 columns in row). Some local variable used to keep calculated size value might a good idea here, then return would easily fit before function name and be consistent with the rest of the code.

ktrzcinx (changes requested):
Please apply changes for r_manifest_sign_v1_5 also.

lgirdwood (commented):
Btw, does the caller provide any context for the error or does it make more sense here to provide context in the local signing function. i.e.
```c
ret = pkcs();
if (ret < 0)
    printf("some meaningfull error messages related to this pkcs func usage");
```

marc-hb:
There is already a "signature file not found" message somewhere, but then rimage signed nothing and returned success. Fixed by this PR.

src/pkcs1_5.c
marc-hb: Done, thanks for the suggestion! 2 commits now, see updated github description and commit messages

src/pkcs1_5.c
```diff
@@ -244,23 +244,25 @@ int ri_manifest_sign_v1_5(struct image *image)
...
-				sizeof(struct css_header_v1_8) -
-				(MAN_RSA_KEY_MODULUS_LEN +
-				 MAN_RSA_KEY_EXPONENT_LEN +
-				 MAN_RSA_SIGNATURE_LEN),
-				(void *)man + MAN_SIG_PKG_OFFSET_V1_8,
-				(man->css.size - man->css.header_len)
-				 * sizeof(uint32_t));
-	return 0;
+	char *const data1 = (char *)man + MAN_CSS_HDR_OFFSET;
+	unsigned const size1 =
```
ktrzcinx, line 258: `uint32_t` would be shorter and consistent with the rest of a code

marc-hb: I have no strong opinion on `uint32_t` versus `unsigned` here but did you see that  `uint32_t` would be _inconsistent_ with the functions this is passed to? I used `unsigned` here because the only purpose of the constant (optimized away by the compiler) is to give a name to the function parameter so it made sense to me to also give the reader the type of the function parameter at the same time.

BTW this line already involves 2 integer types: `size_t` and `unsigned`. Using `uint32_t` would add a third one.

marc-hb:
The checkpatch warning is a bug in checkpatch, probably because it's not running in a kernel clone. The warning normally looks like `WARNING: struct fubar should normally be const` where `fubar` is part of the `scripts/const_structs.checkpatch` list. 
For `rimage` there is no `scripts/const_structs.checkpatch` list and you can see the struct name is missing from the error message printed:
 `WARNING: struct    should normally be const`

ktrzcinx (approved):

lgirdwood (approved):

lgirdwood:
@marc-hb checkpatch should appear as a seperate CI test under checks. Thta way it's obvious. Is this something you or @xiulipan can fix ?

marc-hb:
I got familiar with Travis now but that would be a github change, so no, I unfortunately don't know how to do that. I also don't remember any github repo where we show two different Travis checks on two different lines.

xiulipan:
@lgirdwood @marc-hb Travis do not support this kind of reequipments. But we can try to add another CI for the code scan.

lgirdwood:
@xiulipan thanks, SOF FW has the Checkpatch split from build.

xiulipan:
@lgirdwood @marc-hb maybe remove check patch from Travis?

lgirdwood:
Yes, but it should work the same way as FW where it runs as a separate GH check.

lgirdwood:
> Many good ideas but to be honest that seems like a fair amount of work to save very little time. It's really not the end of the world to open the Travis page and ignore a bogus checkpatch warning now and then, that's a very small part of the time it takes to investigate `rimage` crashes. It's not like there is a lot of activity in this git repo either.

I am very time poor and this more than doubles the review time (and I have driver and FW to review too). We have already done this for FW so why cant it be copied and done here ?

marc-hb:
> We have already done this for FW so why cant it be copied and done here ?

Maybe it can, I don't know. I'm absolutely not opposed to it but I feel like there is a "bigger problem" here.

> I am very time poor and this more than doubles the review time 

I wasn't even considering you. It is not the job of top-level maintainers to open and inspect CI failures. It is the job of submitters to either fix them or justify them in the comments. If they don't then too bad for them, their PR gets delayed until they do.

> (and I have driver and FW to review too).

This means more reviewers are needed. I just reviewed some rimage PRs but that's only because I noticed them  by sheer chance because I was fixing some rimage stuff myself too, no one ever added me to them. Adding more reviewers can be done either automatically using a github team (I just configured this in sof-test) or manually or both.

PR #30, opened 2020-10-15
Keep available manifest parsers in list, move config version parsing to separate function

ktrzcinx:
Improve code readability and allow to easily expand parsers list in future.

### DIFF
--- a/src/adsp_config.c
+++ b/src/adsp_config.c
@@ -1562,13 +1562,35 @@ static int parse_version(toml_table_t *toml, int64_t version[2])
 	return 0;
 }
 
-static inline bool check_config_version(int major, int minor, const int64_t *version)
+struct config_parser {
+	int major;
+	int minor;
+	int (*parse)(const toml_table_t *toml, struct adsp *out, bool verbose);
+};
+
+static const struct config_parser *find_config_parser(int64_t version[2])
 {
-	return version[0] == major && version[1] == minor;
+	/* list of supported configuration version with handler to parser */
+	static const struct config_parser parsers[] = {
+		{1, 0, parse_adsp_config_v1_0},
+		{1, 5, parse_adsp_config_v1_5},
+		{1, 8, parse_adsp_config_v1_8},
+		{2, 5, parse_adsp_config_v2_5},
+	};
+	int i;
+
+	for (i = 0; i < ARRAY_SIZE(parsers); ++i) {
+		if (parsers[i].major == version[0] &&
+		    parsers[i].minor == version[1]) {
+			return &parsers[i];
+		}
+	}
+	return NULL;
 }
 
 static int adsp_parse_config_fd(FILE *fd, struct adsp *out, bool verbose)
 {
+	const struct config_parser *parser;
 	int64_t manifest_version[2];
 	toml_table_t *toml;
 	char errbuf[256];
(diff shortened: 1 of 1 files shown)


src/adsp_config.c
```diff
@@ -1575,35 +1589,21 @@ static int adsp_parse_config_fd(FILE *fd, struct adsp *out, bool verbose)
...
-		ret = parse_adsp_config_v1_8(toml, out, verbose);
-	} else if (check_config_version(2, 5, manifest_version)) {
-		ret = parse_adsp_config_v2_5(toml, out, verbose);
-	} else {
-		ret = log_err(-EINVAL, "error: Unsupported config version %d.%d\n",
-			      manifest_version[0], manifest_version[1]);
-		goto error;
-	}
+	ret = log_err(-EINVAL, "error: Unsupported config version %d.%d\n",
+		      manifest_version[0], manifest_version[1]);
```
marc-hb, line 1604: This really obfuscates the `ret` value to save extremely little code.

src/adsp_config.c
```diff
@@ -1575,35 +1589,21 @@ static int adsp_parse_config_fd(FILE *fd, struct adsp *out, bool verbose)
...
 
-	/* parse "version" array elements */
-	for (i = 0; i < ARRAY_SIZE(manifest_version); ++i) {
-		raw = toml_raw_at(arr, i);
-		if (ret < 0)
-			ret = err_key_parse("version", NULL);
-		ret = toml_rtoi(raw, &manifest_version[i]);
-		if (ret < 0)
-			ret = err_key_parse("version", "can't convert element to integer");
+	/* look for parser related with manifest version and run it */
```
marc-hb, line 1594: s/related/compatible/ ?

marc-hb (commented):
The first commit has two copies of the same `/* parse "version" array elements */` code and computes manifest_version[] twice. It's removed in the second commit but the state between the two commits is weird/inconsistent. I guess it's just a git "fumble"?

src/adsp_config.c
ktrzcinx: changed

src/adsp_config.c
ktrzcinx: Ok, so I extracted `find_config_parser()` function, then error flow is everywhere inside `if` statement and success flow is outside `if`

ktrzcinx:
Checkpatch failed, but actually I don't know why, perhaps some checkpatch issue?

marc-hb:
> Checkpatch failed, but actually I don't know why, perhaps some checkpatch issue?

I think so, same bug I just described in https://github.com/thesofproject/rimage/pull/29#issuecomment-710748554

Looks like this bug is going to trigger every time there is a `struct` :-( Maybe it has been fixed upstream?

src/adsp_config.c
```diff
@@ -1532,6 +1532,33 @@ static int parse_adsp_config_v2_5(const toml_table_t *toml, struct adsp *out,
 	return 0;
 }
 
+static int parse_version(toml_table_t *toml, int64_t version[2])
```
lgirdwood, line 1535: What constitutes a correct version string here, can you say in an inline comment.

src/adsp_config.c
```diff
@@ -1559,13 +1559,35 @@ static int parse_version(toml_table_t *toml, int64_t version[2])
 	return 0;
 }
 
-static inline bool check_config_version(int major, int minor, const int64_t *version)
+struct parser {
```
lgirdwood, line 1562: version_parser ? It seems specific for parsing versions only ?

lgirdwood (changes requested):
Just some more comments needed.

src/adsp_config.c
ktrzcinx: no, it's parser for whole configuration file, version is used only to select right parser

src/adsp_config.c
ktrzcinx: version is not stored in string, it is in toml array of integers. I've added this information as comment.

src/adsp_config.c
ktrzcinx: renamed to `config_parser`

lgirdwood (approved):

lgirdwood:
CI showing checkpatch errors.

PR #31, opened 2020-10-15
Assert valid memory region before hash calculation

ktrzcinx:


ktrzcinx:
I start this branch from outdated master, on current one, there already is assert for number of given elf files. Assertion before hash calculation is still valid change, so I won't close this PR.

src/hash.c
```diff
@@ -82,6 +83,7 @@ void module_sha_complete(struct image *image, uint8_t *hash)
 void ri_sha256(struct image *image, unsigned int offset, unsigned int size,
 	       uint8_t *hash)
 {
+	assert(size < (int64_t)image->adsp->image_size - offset);
```
marc-hb, line 86: When possible and when it makes sense, I think it's more readable to avoid subtractions:
```shell
assert((uint64_t) offset + size < image->adsp->image_size)
```
This order is also closer to the argument order of the `module_sha_update()` call immediately below.

marc-hb (commented):
Are you sure it's not `<=`? `<` seems too restrictive, consider for instance the case when `offset == 0`.

src/hash.c
ktrzcinx: I focus on negative 'size', it why there was little non-standard order, but you're right, I can just extend temporary value to uin64_t to achieve it. Moreover changed condition to `<=`

marc-hb (approved):

lgirdwood (approved):

PR #37, opened 2020-12-04
update the config for some similar DSP config

xiulipan:
- merge JSL and ICL conifg
- use TGL config for TGL-H


config/tgl-h.toml
```diff
@@ -0,0 +1 @@
+tgl.toml
```
lgirdwood, line 1: Should there be a newline here. It's showing up in red bold here ??

xiulipan: This is a soft-link and git could not show correctly here.

lgirdwood (approved):

PR #40, opened 2021-04-12
rimage: add cavs fw layout support

RanderWang:
SOF will support entended manifest and modules. Now
we will have another config tgl-cavs for tgl.


config/tgl-cavs.toml
```diff
@@ -0,0 +1,196 @@
...
+
+[adsp_file]
+[[adsp_file.comp]]
+base_offset = "0x2000"
+
+[fw_desc.header]
+name = "ADSPFW"
+load_offset = "0x30000"
+
+[module]
```
lgirdwood, line 55: It would be a lot better if we could reuse the existing machine definitions and #include the extended manifest e.g.
```
[module]
file = "BRNGUP.toml"
file = "MIXIN.toml"
```

config/tgl-cavs.toml
```diff
@@ -0,0 +1,196 @@
...
+	name = "MIXIN"
+	uuid = "39656EB2-3B71-4049-8D3F-F92CD5C43C09"
+	affinity_mask = "0x1"
+	instance_count = "30"
+	domain_types = "0"
+	load_type = "0"
+	module_type = "1"
+	auto_start = "0"
+	sched_caps = [1, 0x00008000]
+	pin = [0, 0, 0xfeef, 0xc, 0x8, 0x45ff,
```
lgirdwood, line 87: Can you define the pin array as a comment, it looks like an array of pins ?
```
#pin format  [id, flagX, mask, rate, format]
pin = [0, 0, 0xfeef, 0xc, 0x8, 0x45ff,
```

config/tgl-cavs.toml
```diff
@@ -0,0 +1,196 @@
...
+		   1, 0, 0xfeef, 0xc, 0x8, 0x1ff,
+		   1, 0, 0xfeef, 0xc, 0x8, 0x1ff,
+		   1, 0, 0xfeef, 0xc, 0x8, 0x1ff,
+		   1, 0, 0xfeef, 0xc, 0x8, 0x1ff]
+	mod_cfg = [ 0, 0, 0, 0, 296, 644000, 45, 60, 0, 0, 0,
```
lgirdwood, line 92: ditto for mod_cfg - The intention is that anyone should be able to understand these and edit them (until automatic generation is ready).

src/include/rimage/cavs/cavs_ext_manifest.h
```diff
@@ -0,0 +1,249 @@
...
+//      SchedulingCapability[]
+//      PinDescr[]
+
+/** ExtendedManifestHeader id $AE1 */
+#define EXTENDED_MANIFEST_MAGIC_HEADER_ID	0x31454124
+
+#define FW_MAX_EXT_MODULE_NUM	32
+
+struct uuid_t {
+ uint32_t d0;
```
lgirdwood, line 30: alignment

src/include/rimage/cavs/cavs_ext_manifest.h
```diff
@@ -0,0 +1,249 @@
...
+            uint16_t d2;
+            uint8_t d3;
+            uint8_t d4;
+            uint8_t d5;
+            uint8_t d6;
+            uint8_t d7;
+            uint8_t d8;
+            uint8_t d9;
+            uint8_t d10;
+};
```
lgirdwood, line 41: packed ?

src/include/rimage/cavs/cavs_ext_manifest.h
```diff
@@ -0,0 +1,249 @@
...
+	uint32_t            ext_module_config_length;       // sizeof(fw_ext_mod_config_header) in bytes
+	uint32_t            guid[4];                        // Module GUID
+	uint16_t            version_major;                  // Module version
+	uint16_t            version_minor;                  // Module version
+	uint16_t            version_hotfix;                 // Module version
+	uint16_t            version_build;                  // Module version
+	enum mod_type module_type;
+	uint32_t            init_settings_min_size;         // Minimum size of initialization settings (in bytes)
+	uint16_t            num_scheduling_capabilities;    // number scheduling capabilities supported by the module
+	uint16_t            num_pin_entries;                // Number of Pin (inputs + ouptuts)
```
lgirdwood, line 233: Can we make sure all comments are doxygen style and that we pack the data. The alignment looks off in places too.

src/manifest.c
```diff
@@ -1204,6 +1237,85 @@ int man_write_fw_meu_v2_5(struct image *image)
...
+	/* firmware and build version */
+	desc->header.major_version = image->fw_ver_major;
+	desc->header.minor_version = image->fw_ver_minor;
+	desc->header.build_version = image->fw_ver_build;
+
+	/* create each module */
+	man_create_modules(image, desc, FILE_TEXT_OFFSET_V1_8);
+	man_create_modules_v2_6(image, desc);
+
+	fprintf(stdout, "Firmware completing manifest v2.5\n");
```
lgirdwood, line 1285: 2.6

lgirdwood (changes requested):

src/include/rimage/cavs/cavs_ext_manifest.h
RanderWang: no, I first added packed attribute but fw_parser can't decode it correctly, so I removed it.

lgirdwood: hmm, these are ABI so should be packed. Best to do a comparison with the other binaries and codebase.

config/tgl-cavs.toml
RanderWang: let me check toml

RanderWang: And we still need to reform existing machine definitions if toml is used since the definition format is not compatible with toml

RanderWang: Another way without modifying existing machine definitions: use our own code to decode these config files. I first implemented some functions to achieve this, but at last I decided to use current toml tools.  @lgirdwood what's your idea ?

src/include/rimage/cavs/cavs_ext_manifest.h
RanderWang: yes, thanks!

RanderWang:
sorry,  I will push it to main branch , not master

PR #42, opened 2021-04-13
rimage: add cavs fw layout support

RanderWang:
Add support of cavs extended manifest and modules. Now
we will have another config tgl-cavs for tgl.


src/adsp_config.c
```diff
@@ -1915,6 +2268,72 @@ static int parse_adsp_config_v2_5(const toml_table_t *toml, struct adsp *out,
 	return 0;
 }
 
+static int parse_adsp_config_v2_6(const toml_table_t *toml, struct adsp *out,
```
lgirdwood, line 2271: Whats the difference here with v2.5 ? My understanding is that only the ext manifest is different ?

src/include/rimage/cavs/cavs_ext_manifest.h
```diff
@@ -0,0 +1,266 @@
...
+ *      PinDescr[]
+ */
+
+/* ExtendedManifestHeader id $AE1 */
+#define EXTENDED_MANIFEST_MAGIC_HEADER_ID	0x31454124
+
+#define FW_MAX_EXT_MODULE_NUM	32
+
+struct uuid_t {
+ uint32_t d0;
```
lgirdwood, line 31: alignment

src/include/rimage/cavs/cavs_ext_manifest.h
```diff
@@ -0,0 +1,266 @@
...
+	struct mod_scheduling_caps sched_caps;
+	struct fw_pin_description *pin_desc;
+};
+
+struct fw_ext_manifest
+{
+	struct fw_ext_man_header header;
+	int mod_conf_count;
+	struct fw_ext_mod_config ext_mod_config_array[FW_MAX_EXT_MODULE_NUM];
+};
```
lgirdwood, line 265: These will all need packed.

src/include/rimage/manifest.h
```diff
@@ -127,6 +128,15 @@ struct fw_image_manifest_v2_5 {
...
 } __attribute__((packed));
 
+struct fw_image_manifest_v2_6 {
+	struct fw_image_manifest_v2_5 man_v2_5;
+	struct fw_ext_manifest ext_man;
+	int mod_cfg_count;
+	struct sof_man_mod_config *mod_cfg;
+	int mod_man_count;
+	struct sof_man_module *mod_man;
+};
```
lgirdwood, line 138: This looks very similar to v2.5 ? Can we reuse ? and set the new 2.6 fields to 0 in 2.5 ?

lgirdwood (changes requested):
It looks like the v2.6 contain v2.5 but has following module data. We should just add the modules to v2.5 and set them to 0 when not used.

src/include/rimage/cavs/cavs_ext_manifest.h
RanderWang: @lgirdwood  this structure is not defined by fw spec, just by me to save module information to build manifest later, so we don't need to pack it.

RanderWang: For above structures, it can't be packed since adsp_fw_parser can't decode it correctly. And other drivers follow adsp_fw_parser, so we don't packed attribute to be compatible with them

src/include/rimage/manifest.h
RanderWang: yes, please check my new PR

RanderWang:
@lgirdwood  I merged 2_6 to 2_5. The major gap is that we don't support modules in extended manifest and module configs except bring-up & base module. I check whether there is module defined in platform config for cavs and create cavs layout manifest if yes

src/include/rimage/cavs/cavs_ext_manifest.h
lgirdwood: manifest.h is only for structure that are in the manifest. Internal only structures should go in internal headers.

This needs to be packed otherwise it will build differently with 64bit and 32bit compilers. So instead of using an ```int``` use a int32_t or int64_t to align with the parser..

src/rimage.c
```diff
@@ -209,7 +209,11 @@ int main(int argc, char *argv[])
 
 	/* build extended manifest */
 	if (use_ext_man) {
-		ret = ext_man_write(&image);
+		if (image.adsp->wirte_firmware_ext_man)
```
lgirdwood, line 212: typo - should be ```write```

lgirdwood (changes requested):
It all needs packed with correct type sizes if its in the manifest.

src/rimage.c
RanderWang: ...... thanks

src/include/rimage/cavs/cavs_ext_manifest.h
RanderWang: thanks!

RanderWang:
@lgirdwood  update my PR, thanks!

src/include/rimage/cavs/cavs_ext_manifest.h
```diff
@@ -0,0 +1,253 @@
...
+		uint16_t all : 1;
+	} r;
+};
+
+struct mod_scheduling_caps
+{
+	/* scheduling period in Samples (sample groups) (note: 1 Sample = 1 sample per channel) */
+	uint16_t    frame_length;
+	union mod_multiples   multiples_supported;
+};
```
lgirdwood, line 73: All structure in here need packed with correct type sizes optional padding to align with parser.

lgirdwood (changes requested):
We do need to pack and use correct type sizes for any data written to file.

src/include/rimage/cavs/cavs_ext_manifest.h
RanderWang: I will check it with adsp_fw_parser.  And the sizeof mod_multiples is also uint16.

RanderWang:
@lgirdwood  I tested every structure in cavs_ext_manifest.h with packed attribute.  Only one structure can't be packed: struct fw_pin_description.  adsp_fw_parser can't decode it correctly. I debugged with gdb and found structure data is out of order.   I added packed attribute for all other structure

lgirdwood:
> @lgirdwood I tested every structure in cavs_ext_manifest.h with packed attribute. Only one structure can't be packed: struct fw_pin_description. adsp_fw_parser can't decode it correctly. I debugged with gdb and found structure data is out of order. I added packed attribute for all other structure

The unions also need packed. I suspect one of the uint16_t unions is really as uint32_t.

RanderWang:
> > @lgirdwood I tested every structure in cavs_ext_manifest.h with packed attribute. Only one structure can't be packed: struct fw_pin_description. adsp_fw_parser can't decode it correctly. I debugged with gdb and found structure data is out of order. I added packed attribute for all other structure
> 
> The unions also need packed. I suspect one of the uint16_t unions is really as uint32_t.

yes,  sizeof(packed) is different with sizeof(non-packed).   I did a experiment , it is what you suspected:  uint16_t in structure occupies 32bit without packed attribute. I think we have to match with driver usage.

src/include/rimage/cavs/cavs_ext_manifest.h
```diff
@@ -0,0 +1,253 @@
...
+		uint16_t x9 : 1;
+		uint16_t x10 : 1;
+		uint16_t x11 : 1;
+		uint16_t x12 : 1;
+		uint16_t x13 : 1;
+		uint16_t x14 : 1;
+		uint16_t x15 : 1;
+		uint16_t all : 1;
+	} r;
+};
```
lgirdwood, line 66: This needs packed. If parser fails with packing then ul is a uint32_t. Ditto for other unions.

lgirdwood (changes requested):
All structs and unions must be packed. It's failing parser as some uint16_t unions are really uint32_t.

RanderWang:
update PR. Thanks! the following structure can't be packed. the layout of data is different.  For example,  for no-packed containers  mod_sample_containers.ul = 0x8,  packed mod_sample_containers.ul will be 0x80000;  the definition of this structure is very ugly, the reverved should use 27 bits.

````C
union mod_sample_containers
{
	uint32_t ul;
	struct
	{
		uint16_t bits_8 : 1;
		uint16_t bits_16 : 1;
		uint16_t bits_24 : 1;
		uint16_t bits_32 : 1;
		uint16_t bits_64 : 1;
		uint16_t reserved : 11;
	} r;
} ;
`````

lgirdwood:
They all have to be packed. This is a BUG in the original code, the packed C structure must be aligned with the binary otherwise this will break at some point when any CC decides to format the data differently. 

However, the struct can be fixed to have the data correctly formatted. e.g. have you trued setting all the bits_ as uint32_t ?? or have you tried more reserved paddding (like below)

> ```c
> union mod_sample_containers
> {
> 	uint32_t ul;
> 	struct
> 	{
 reserved1 : 16; // padding here.
> 		uint16_t bits_8 : 1;
> 		uint16_t bits_16 : 1;
> 		uint16_t bits_24 : 1;
> 		uint16_t bits_32 : 1;
> 		uint16_t bits_64 : 1;
> 		uint16_t reserved : 11;
> 	} r;
> } ;
> ```

RanderWang:
> They all have to be packed. This is a BUG in the original code, the packed C structure must be aligned with the binary otherwise this will break at some point when any CC decides to format the data differently.
> 
> However, the struct can be fixed to have the data correctly formatted. e.g. have you trued setting all the bits_ as uint32_t ?? or have you tried more reserved paddding (like below)
> 
> > ```c
> > union mod_sample_containers
> > {
> > 	uint32_t ul;
> > 	struct
> > 	{
> > ```
> 
> reserved1 : 16; // padding here.
> 
> > ```
> >   uint16_t bits_8 : 1;
> >   uint16_t bits_16 : 1;
> >   uint16_t bits_24 : 1;
> >   uint16_t bits_32 : 1;
> >   uint16_t bits_64 : 1;
> >   uint16_t reserved : 11;
> > ```
> > 
> > 
> > } r;
> > } ;

We can fix it for SOF, but how about close source driver ?  It can't work with our FW( I checked our FW with close source parser tool), will we submit a patch to fix it ?

lgirdwood:
> > They all have to be packed. This is a BUG in the original code, the packed C structure must be aligned with the binary otherwise this will break at some point when any CC decides to format the data differently.
> > However, the struct can be fixed to have the data correctly formatted. e.g. have you trued setting all the bits_ as uint32_t ?? or have you tried more reserved paddding (like below)
> > > ```c
> > > union mod_sample_containers
> > > {
> > > 	uint32_t ul;
> > > 	struct
> > > 	{
> > > ```
> > 
> > 
> > reserved1 : 16; // padding here.
> > > ```
> > >   uint16_t bits_8 : 1;
> > >   uint16_t bits_16 : 1;
> > >   uint16_t bits_24 : 1;
> > >   uint16_t bits_32 : 1;
> > >   uint16_t bits_64 : 1;
> > >   uint16_t reserved : 11;
> > > ```
> > > 
> > > 
> > > } r;
> > > } ;
> 
> We can fix it for SOF, but how about close source driver ? It can't work with our FW( I checked our FW with close source parser tool), will we submit a patch to fix it ?

Sorry, what I mean is the structure must be packed (and reworked to align with the ABI used by parser). This way it will be consistently always the same no matter which compiler is used. So what needs done is
1) Pack the struct/union.
2) align it with ABI by either changing bits to ```c uint32_t bits_x : x``` or by adding more reserved padding.

RanderWang:
update my PR to pack all ABI structures.   Add reserved padding bit to make parser happy.   My original idea was to use default definition without any change to align with them.  Now I realize that we don't need to use default definition and keep compatible with parser with padding bits.

lgirdwood (approved):

lgirdwood:
Can you take a look and fix the check patch errors. Most of the warnings and checks can be ignored.

RanderWang:
@lgirdwood  I updated PR to fix all the errors. The others are warnings, unreasonable warning.

PR #43, opened 2021-04-13
rimage: add support to strip signature in signed fw

RanderWang:
CI has a requirment to compare fw content without any
signature to check whether there is no change between
two fw binaries. This patch will strip signature in
signed fw and store it to no_sig_ + fw name.

please check https://github.com/thesofproject/rimage/issues/41

tested on APL & TGL

### DIFF
--- a/src/rimage.c
+++ b/src/rimage.c
@@ -47,7 +47,7 @@ int main(int argc, char *argv[])
 
 	image.xcc_mod_offset = DEFAULT_XCC_MOD_OFFSET;
 
-	while ((opt = getopt(argc, argv, "ho:va:s:k:ri:x:f:b:ec:y:")) != -1) {
+	while ((opt = getopt(argc, argv, "ho:va:s:k:ri:x:f:b:ec:y:p:")) != -1) {
 		switch (opt) {
 		case 'o':
 			image.out_file = optarg;
@@ -88,6 +88,9 @@ int main(int argc, char *argv[])
 		case 'y':
 			image.verify_file = optarg;
 			break;
+		case 'p':
+			image.strip_file = optarg;
+			break;
 		case 'h':
 			usage(argv[0]);
 			break;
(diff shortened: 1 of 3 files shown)


src/rimage.c
```diff
@@ -88,6 +88,9 @@ int main(int argc, char *argv[])
 		case 'y':
 			image.verify_file = optarg;
 			break;
+		case 'p':
+			image.verify_file = optarg;
```
lgirdwood, line 92: Why is this variable being reused for a different purpose ?

lgirdwood (commented):
Would it not be better to just bypass the signing code (and hence the hashing and signing content will be 0).

RanderWang:
> Would it not be better to just bypass the signing code (and hence the hashing and signing content will be 0).

you mean we bypass signing code when building FW ? we don't need to do it since we build a FW without any signature then add signature by rimage. But this is not fit for marc. This is his email:
"
  I'm not sure you understand the build reproducibility idea. The idea is that for instance Curtis from Google does not trust the releases in sof-bin because he suspects they have been tampered with for some reason. For instance because someone hacked our Jenkins server, the one that has all the security vulnerabilities and alerts on its home page. This type of "supply chain" attack is more and more common. On the other hand, Curtis trusts the SOF sources because he has been scanning and monitoring them very closely. So he tries to rebuild the release himself and wants to compare it to the release in sof-bin. However the signature is now randomized with a salt, so the .ri files are always different even when their content is actually the same. "Un-rimaging" solves that problem, hence my question below.

Marc

PS: I'm also using build reproducibility locally to make sure some code changes don't affect the compiler output, that's super convenient because it means I don't need to run any test when I make such a change. In my workspace I have hardcoded the rimage date and salt but I obviously cannot submit that.
"

RanderWang:
@marc-hb  do you have any comments to me and Liam ?

src/rimage.c
RanderWang: thanks, I will refine it.

lgirdwood:
Manifest already includes reproducible hashes of the content. Lets use them.

PR #44, opened 2021-04-26
ext_man: Fix memory leak after manifest build

ktrzcinx:
sec_buffer is allocated inside elf_read_section() and as local
variable it's deleted after ext_man_build() exit, so free()
should be called here.

### DIFF
--- a/src/ext_manifest.c
+++ b/src/ext_manifest.c
@@ -97,7 +97,7 @@ static int ext_man_build(const struct module *module,
 		fprintf(stderr,
 			"error: failed to read %s section content, code %d\n",
 			EXT_MAN_DATA_SECTION, ret);
-		goto out;
+		return ret;
 	}
 	ret = 0;
 
@@ -128,6 +128,7 @@ static int ext_man_build(const struct module *module,
 	*dst_buff = (struct ext_man_header *)buffer;
 
 out:
+	free(sec_buffer);
 	return ret;
 }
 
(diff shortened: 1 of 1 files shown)


src/ext_manifest.c
```diff
@@ -97,7 +97,7 @@ static int ext_man_build(const struct module *module,
 		fprintf(stderr,
 			"error: failed to read %s section content, code %d\n",
 			EXT_MAN_DATA_SECTION, ret);
-		goto out;
+		return ret;
```
paulstelian97, line 100: This change is extremely weird given the commit message.

paulstelian97 (changes requested):

src/ext_manifest.c
paulstelian97: Also unneeded, since you'd call free(NULL) which is completely safe to do. So unneeded even if correct actually.

lgirdwood: The code looks correct (functionally the same if we have the ```goto``` or the ```return``` here), I agree I'm a little confused by the commit message.

lgirdwood (changes requested):
Code is fine - I would just reword the commit message. sec_buffer was not being freed before and is freed now.

lgirdwood:
@ktrzcinx was this pushed directly by accident ? The commit message needed some work.

ktrzcinx:
> @ktrzcinx was this pushed directly by accident ? The commit message needed some work.

Yes, it was accident - I have reinstalled OS and a few of my configs has gone so now I've made some mess.
Although I applied yours suggestions and force pushed to rimage master branch (with success).
Moreover 'master' shows me as default one instead of 'main'.

PR#45 should be on proper branches and with revised commit version

lgirdwood:
No problem, I also noticed this was on master to which I thought was deleted. Deleted master now.

PR #46, opened 2021-05-03
Cppcheck fixes but its for rimage

cujomalainey:


cujomalainey:
@lgirdwood do you know why the workflows are not being recognized? Also @marc-hb I have a commit to update this to actions and off travis

marc-hb:
Please try to copy/paste one of these two files into https://github.com/thesofproject/rimage/actions/new

For the main repo I think I remember nothing happened until I started using the web interface.

cujomalainey:
@marc-hb i think i figured it out, likely wont trigger new files when being pulled from an external fork for security reasons, since it ran fine in my repo

cujomalainey:
nevermind, added and deleted a commit and now it all works. i blame cosmic radiation

marc-hb:
> i blame cosmic radiation

I did not complete it but I did initiate the addition of a github action from the web interface... blame me?

cujomalainey:
> > i blame cosmic radiation
> 
> I did not complete it but I did initiate the addition of a github action from the web interface... blame me?

Seems to be a common issue https://github.community/t/workflow-dispatch-workflow-not-showing-in-actions-tab/130088/25

.github/workflows/cppcheck.yml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  cppcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        # From time to time this will catch a git tag and change SOF_VERSION
+        with: {fetch-depth: 50, submodules: recursive}
+
+      - name: apt install cppcheck
+        run: sudo apt-get -y install cppcheck
```
marc-hb, line 26: You need `apt update` before any install unfortunately: https://github.com/actions/virtual-environments/issues/2924

.github/workflows/cppcheck.yml
```diff
@@ -0,0 +1,31 @@
...
+# yamllint disable-line rule:truthy
+on: [pull_request, push]
+
+jobs:
+  cppcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        # From time to time this will catch a git tag and change SOF_VERSION
+        with: {fetch-depth: 50, submodules: recursive}
```
marc-hb, line 23: There is no SOF_VERSION in this repo. You could replace SOF_VERSION with `git describe`.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,29 @@
...
+# yamllint disable-line rule:truthy
+on: [pull_request]
+
+jobs:
+  yamllint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        # From time to time this will catch a git tag and change SOF_VERSION
+        with: {fetch-depth: 50, submodules: recursive}
```
marc-hb, line 23: There is no SOF_VERSION in this repo. You could replace SOF_VERSION with `git describe`.

src/adsp_config.c
```diff
@@ -1867,7 +1867,7 @@ static int parse_mod_config(const toml_table_t *mod_entry, struct parse_ctx *ctx
 static void parse_uuid(char *buf, uint8_t *uuid)
 {
 	struct uuid_t id;
-	uint32_t d[9];
+	uint32_t d[10];
```
marc-hb, line 1870: Ouch.

src/ext_manifest.c
```diff
@@ -131,6 +131,7 @@ static int ext_man_build(const struct module *module,
 
 out:
 	free(sec_buffer);
+	/* cppcheck-suppress memleak */
```
marc-hb, line 134: I don't think that's correct, I think cppcheck found a real bug. I believe a `free(buffer)` is missing here.

Now I'm not too worried about memory leaks in a tool like `rimage` but if it's possible to fix warnings by fixing bugs then we might as well.

dcpleung: I think it's complaining about the `calloc()` at line 118. `buffer` is not being freed here, hence the warning. I wonder if you can tell cppcheck to ignore this particular variable instead of ignoring all memleak. Or maybe the calling of this functions never frees `dst_buff`?

marc-hb: Sorry I didn't read the whole function and missed `dst_buffer`.

You could also get rid of `buffer`, it's not very useful. I tested this and it works:
```diff
 	/* fill buffer with ext_man and section content */
-	memcpy(buffer, &ext_man, ext_man.header_size);
 	offset = ext_man.header_size;
+	memcpy(*dst_buff, &ext_man, offset);
 
-	memcpy(&buffer[offset], sec_buffer, section->size);
-
-	*dst_buff = (struct ext_man_header *)buffer;
+	memcpy((char *)*dst_buff + offset, sec_buffer, section->size);
 
 out:
 	free(sec_buffer);
```

cujomalainey: I rather refactor a little to silence the false positive than put down a suppressor. I will try this out. I also filed a bug upstream against this false positive

src/adsp_config.c
cujomalainey: Yea, that one hurt when i found it

src/ext_manifest.c
cujomalainey: @dcpleung i checked the caller and it handles it fine, cppcheck just fails to understand that dst_buff passes the memory out of scope and therefore is not a leak

lgirdwood (commented):
We can get this in rc2.

cujomalainey:
Sorry for the delay, got caught up in internal stuff

cujomalainey:
Travis is failing because checkpatch is mad that im using `memcpy` vs `memcpy_s`. Using the latter will require more surgery since size is not passed into the function. @lgirdwood what are your thoughts here

src/ext_manifest.c
```diff
@@ -114,20 +113,11 @@ static int ext_man_build(const struct module *module,
...
-	/* alloc buffer for ext_man */
-	buffer = calloc(1, ext_man.full_size);
-	if (!buffer) {
-		ret = -ENOMEM;
-		goto out;
-	}
-
 	/* fill buffer with ext_man and section content */
-	memcpy(buffer, &ext_man, ext_man.header_size);
+	memcpy(*dst_buff, &ext_man, ext_man.header_size);
```
fredoh9, line 117: set aside memcpy_s(), this doesn't seems right. `*dst_buff` has only space for a pointer. How can we sure there is enough size for ext_man.header_size? I think, this is good reason that we need to move toward memcpy_s()

marc-hb: You can't drop the `calloc`, you still need something like `*dst_buffer = calloc(...`

marc-hb (changes requested):
See below

src/ext_manifest.c
lgirdwood: Yeah, it does not look like the dest buf is being allocated.

lgirdwood (commented):
Its possible the testing used the version of rimage in $PATH. IIRC cmake check in $PATH first when it's looking for tools.

src/ext_manifest.c
cujomalainey: Yes, apologies, I did the CI changes second and forgot to test rimage. Will fix

fredoh9 (approved):
This is great fixes!

cujomalainey:
FWIW i tried to update to `memcpy_s` but the docker container we use right now doesn't have the definition so it breaks it

marc-hb (approved):
Rimage commit 740b5f54c94cffc7 tested in https://github.com/thesofproject/sof/pull/4205 and everything is green (I hate to admit that submodules are very well supported by github)

>  FWIW i tried to update to memcpy_s but the docker container we use right now doesn't have the definition so it breaks it


Does any Linux distro have `memcpy_s()`?

cujomalainey:
> Rimage commit 740b5f54c94cffc7 tested in https://github.com/thesofproject/sof/pull/4205 and everything is green (I hate to admit that submodules are very well supported by github)
> 
> >  FWIW i tried to update to memcpy_s but the docker container we use right now doesn't have the definition so it breaks it
> 
> 
> Does any Linux distro have `memcpy_s()`?

Github does do an amazing job of handling the evil that is submodules.

I'm not sure what is going on with the container. It's Ubuntu 20.04. memcpy_s has been part of gcc since ~4.8 log shows 9.x. I even checked `/use/include/string.h` and the signature was not there.

marc-hb:
Do you have `memcpy_s` on your Linux system and if yes then what distribution is it?

> memcpy_s has been part of gcc since ~4.8 

I think `memcpy` and similar functions are part of libc, not of any compiler. In fact we recently had an issue where the `gcc -O` optimizer was auto-generating calls to these functions that were not even explicit in the source code and that was causing the linker to fail: https://github.com/thesofproject/sof/issues/3880#issuecomment-796537478

In the past the glibc maintainer(s) have expressed very negative opinions about these extensions submitted by Microsoft.

cujomalainey:
> Do you have `memcpy_s` on your Linux system and if yes then what distribution is it?
> 
```
dpkg -S /usr/include/string.h 
libc6-dev:i386, libc6-dev:amd64: /usr/include/string.h
```

Only place I can see it is either in code that xcc compiles (aka SOF) or in one very obscure place I would not expect to be using any headers from. This is a fork of debian.  Compiling a test program in the base OS also fails. So we can't assume its there it seems.

lgirdwood (approved):

lgirdwood:
Travis checkpatch failure, build is good.

PR #49, opened 2021-05-20
.github/ actions: replace travis

cujomalainey:
We removed them from the main repo, lets remove them here too.

Signed-off-by: Curtis Malainey <cujomalainey@chromium.org>


cujomalainey:
WIP, just figured i would share as an FYI

.github/workflows/build.yml
```diff
@@ -0,0 +1,24 @@
...
+        run : docker pull thesofproject/sof && docker tag thesofproject/sof sof
+
+      - name: Init
+        run: echo -e '#!/bin/bash\nmkdir build && cd build && cmake .. && make -Werror -Wall -Wmissing-prototypes -Wimplicit-fallthrough=3 -Wpointer-arith' > build.sh && chmod +x build.sh
+      - run: docker run -i -t -v $(pwd):/home/sof/work/sof.git --user $(id -u) sof ./build.sh
+  checkpatch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: install codespell
+        run: sudo apt-get -y install codespell
```
marc-hb, line 19: Unlike Travis? Github requires `apt-get -update` first to be on the safe side, see https://github.com/actions/virtual-environments/issues/2924

.github/workflows/build.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  build-test:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Pull Docker
+        run : docker pull thesofproject/sof && docker tag thesofproject/sof sof
+
+      - name: Init
+        run: echo -e '#!/bin/bash\nmkdir build && cd build && cmake .. && make -Werror -Wall -Wmissing-prototypes -Wimplicit-fallthrough=3 -Wpointer-arith' > build.sh && chmod +x build.sh
```
marc-hb, line 13: Note you don't need to escape newlines in this case.

Alternatively, feel free to create a new `scripts/build-with-warnings.sh` file. Or later.

.github/workflows/build.yml
```diff
@@ -0,0 +1,24 @@
...
+      - run: docker run -i -t -v $(pwd):/home/sof/work/sof.git --user $(id -u) sof ./build.sh
```
marc-hb, line 14: > build failed: the input device is not a TTY

Please remove docker entirely here. The docker image is useful for the cross-compilation toolchain, otherwise it's pointless. If we can't easily compile rimage on any random Linux distribution then we have a serious problem.

marc-hb (changes requested):
Thanks for doing this!

Is this missing the "checkout" action?

cujomalainey:
> Is this missing the "checkout" action?

very likely, this was a copy paste refactor when I was doing the cppcheck work and haven't touched it in a few weeks, looking to land it soon.

.github/workflows/pull-request.yml
```diff
@@ -23,3 +23,17 @@ jobs:
...
         run: yamllint .github/workflows/*.yml
+  checkpatch:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with: {fetch-depth: 50, submodules: recursive}
+
+      - name: install codespell
+        run: sudo apt update
+      - run: sudo apt install -y codespell
```
marc-hb, line 34: The list of `run` lines looks weird.

.github/workflows/pull-request.yml
```diff
@@ -23,3 +23,17 @@ jobs:
 
       - name: run yamllint
         run: yamllint .github/workflows/*.yml
+  checkpatch:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with: {fetch-depth: 50, submodules: recursive}
+
+      - name: install codespell
```
marc-hb, line 32: codespell was used by checkpatch but now checkpatch is gone?

marc-hb (commented):
Should be named "move travis.yml to github actions" instead.

cujomalainey:
@marc-hb feel free to ignore pushes till i mark ready for review, just running tests

.github/workflows/build.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  build-test:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with: {fetch-depth: 50, submodules: recursive}
+
+      - name: install cmake
+        run: sudo apt update
+      - run: sudo apt install -y cmake
```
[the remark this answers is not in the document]
cujomalainey: Normally we build in the container though which has this for us, this is outside the container, shouldn't we pull it manually?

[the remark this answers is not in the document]
marc-hb: .github/workflows/unit-tests.yml doesn't build in the container and seems to be doing fine. Github has a list of default packages somewhere but it's probably faster to just try :-) CMake is really the standard so it would be very surprising not to include it.

(Toolchain is of course the big reason why we need containers)

cujomalainey:
@marc-hb i am at a loss for why checkpatch is not working properly, locally when i run the same command it performs the check on only my patch as expected, but in CI it seems to check the whole repo, any guesses what is going on?

lgirdwood (commented):
The checkpatch warnings for line length are getting a little old fashioned. Lets keep up with Linux and allow > 80 char limit (we can align to 100 char line length the same as Linux now accepts).

marc-hb:
@cujomalainey getting checkpatch to work with shallow commits is tricky. It looks like you're copied some old checkpatch Action from SOF. You can either copy the more recent one at https://github.com/thesofproject/sof/blob/main/.github/workflows/shallowfetchPRcommits.sh or simply disable shallow clones are run `checkpatch -g` normally, like you would in your workspace.

lgirdwood (approved):

.github/workflows/pull-request.yml
```diff
@@ -23,3 +23,20 @@ jobs:
...
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: install codespell
+        run: sudo apt update && sudo apt install -y codespell
+
+      - name: run codespell
+        run: curl ${{ github.event.pull_request.commits_url }} | jq '.[].sha'
```
marc-hb, line 38: The curl is awkward. Can you try this instead? From https://github.com/actions/runner/issues/342#issuecomment-875823278

```yaml
with:
    COMMIT_RANGE: ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }}

-name : checkpatch
  ./scripts/checkpatch.pl -g "${COMMIT_RANGE}" --codespell ...
```

.github/workflows/pull-request.yml
```diff
@@ -23,3 +23,20 @@ jobs:
 
       - name: run yamllint
         run: yamllint .github/workflows/*.yml
+  checkpatch:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
marc-hb, line 31: It looks like you don't use $ref.

.github/workflows/pull-request.yml
```diff
@@ -23,3 +23,20 @@ jobs:
...
+          fetch-depth: 0
+
+      - name: install codespell
+        run: sudo apt update && sudo apt install -y codespell
+
+      - name: run codespell
```
marc-hb, line 37:
```suggestion
      - name: checkpatch
```

marc-hb (changes requested):

lgirdwood:
@cujomalainey @marc-hb any resolution here ?

cujomalainey:
I have AIs, I just need to find time to do them

.github/workflows/build.yml
```diff
@@ -0,0 +1,19 @@
...
+  build-test:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with: {fetch-depth: 50, submodules: recursive}
+
+      - name: build
+        run: cmake -B build/
+      - run: make -Werror -Wall -Wmissing-prototypes -Wimplicit-fallthrough=3
+             -Wpointer-arith -C build/
```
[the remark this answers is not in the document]
aborisovich: Nobody uses CMAKE_C_FLAGS anymore unless this kind of operation is required.
Use [target_compile_options](https://cmake.org/cmake/help/latest/command/target_compile_options.html) to specify compilation flags per target. Especially now when we are going to interface with Zephyr. You probably do not want to enforce -Wall on Zephyr or other external dependency?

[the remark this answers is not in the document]
marc-hb: You're confusing command line and scripts. Strange for someone who pretends to know how everyone in the world uses CMake (which is impossible considering how lax CMake is).

rimage is a build-time tool whereas Zephyr is cross-compiled. They have absolutely nothing related to each other.

But since you mentioned Zephyr, its CI uses EXTRA_CFLAGS command line interface to add `-Werror`. That does not seem standard.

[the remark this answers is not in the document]
aborisovich: I do not like your tone @marc-hb nor share your opinion. Watch yourself we work in the company that promotes tolerance and mutual respect so please do so. I do not pretend anything. You pasted link to cmake flags that are considered bad practise. I do not understand why you are trying to add those flags here in the run script.
They are present anyway in rimage/CMakeLists.txt:
`target_compile_options(rimage PRIVATE
	-Wall -Werror -Wmissing-prototypes -Wimplicit-fallthrough
)`
Please explain to me why I am wrong skipping that misplaced attitude.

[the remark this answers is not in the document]
cujomalainey: @aborisovich this is not a script for users, its a CI test. anything here will not leak into other projects as nothing else is being built.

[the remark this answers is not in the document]
aborisovich: @cujomalainey Ok. But why are those flags needed if we have them already defined in rimage/CMakeLists.txt?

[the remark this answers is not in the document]
cujomalainey: This was a copy paste conversion from what we had in the original Travis CI script, given this PR is almost 8months old and the travis script is probably from around 2 years ago, I would not be surprised if we no longer need them and they can be dropped.

marc-hb:
@cujomalainey , if you still don't have the time, maybe @aborisovich could take over?

It looks like https://github.com/actions/virtual-environments offers some Windows environments too.

cujomalainey:
@marc-hb works for me, we had a re-org right after I started this and I have not had as much time to work upstream as I have wanted.

aborisovich:
@marc-hb I am starting next week work related to IPC/IDC Zephyr API usage and concept of a refactored build system that would allow seamlessly build SOF+Zephyr on both Linux and Windows for some platforms will be unfinished. I'm drowning in work for now.

marc-hb:
I understand, thx.

On the other hand, I found Github Actions the best to test build system changes.

.github/workflows/pull-request.yml
cujomalainey: this is passed into checkout

cujomalainey: I think the reason I did this is otherwise it was giving me a merge which doesn't work for checkpatch

cujomalainey:
Eh, 12months later and it still rebases cleanly. Apologies for abandoning this

.github/workflows/build.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  build-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with: {fetch-depth: 50, submodules: recursive}
+
+      - name: build
+        run: cmake -B build/
```
marc-hb, line 20:
```suggestion
        run: cmake -B build/ -GNinja
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,21 @@
...
+      - run: make  -C build/
```
marc-hb, line 21:
```suggestion
      - run: cmake  --build build/
```

.github/workflows/build.yml
marc-hb: `make` does `-j1` by default. Ninja is smarter.

.github/workflows/pull-request.yml
marc-hb: Correct: by default `checkout` gives `pull/12355/merge` which is what most people want by default (even when they don't have a clue).


BTW: https://github.com/thesofproject/sof/blob/main/.github/workflows/codestyle.yml

.github/workflows/pull-request.yml
```diff
@@ -23,3 +23,20 @@ jobs:
 
       - name: run yamllint
         run: yamllint .github/workflows/*.yml
+  checkpatch:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
marc-hb, line 29:
```suggestion
      - uses: actions/checkout@v3
```
v2 has started to print warnings, see for instance: 
https://github.com/thesofproject/sof/pull/6720

See, some things did change :-)

.github/workflows/pull-request.yml
cujomalainey: Interesting, ill look into that tomorrow, thanks for the pointer

cujomalainey: Nvm someone did it, any issue if we just use this? https://github.com/marketplace/actions/checkpatch-pl-pr-review

cujomalainey: Nvm someone did it, any issue if we just use this? https://github.com/marketplace/actions/checkpatch-pl-pr-review

marc-hb: I just had a look at it. The code wouldn't pass `shellcheck` but it's short, understandable and uses the right API (i.e.: [NOT `git merge-base`](https://github.com/actions/checkout/issues/552#issuecomment-1344881907)

Worth a try!

lgirdwood:
@cujomalainey good to target for v2.5 ?

cujomalainey:
Can't make any promises, this is pretty low on my priority list unfortunately

cujomalainey:
@marc-hb @lgirdwood all green, I think this is FINALLY ready to go. Then we can remove any travis hooks left.

lgirdwood (approved):

PR #51, opened 2021-07-02
correct Image size for cAVS platforms

keyonjie:
We have 46 bank HP and 1 bank LP SRAM on TGL, change to toml config file
    to reflect that.

    We have 30 bank HP and 1 bank LP SRAM on TGL-H, change to toml config file
    to reflect that.

### DIFF
--- a/config/apl.toml
+++ b/config/apl.toml
@@ -3,7 +3,7 @@ version = [1, 8]
 [adsp]
 name = "apl"
 machine_id = 5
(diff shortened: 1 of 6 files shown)


lgirdwood (approved):

marc-hb:
To keep the copy/paste/diverge under vague control it would have been good to mention ICL in the JSL file etc.

PR #53, opened 2021-07-20
config: tgl-cavs: change image_size to the real SRAM size

RanderWang:
Apply "config: tgl: change image_size to the real SRAM size"
(9e50a02f1b) to tgl-cavs


keyonjie (approved):
looks good, thanks.

PR #59, opened 2021-08-25
adsp_config: fix potential overflow in strtol()

mrajwa:
This patch fixes the potential overflow which may happen during
conversion of signed long. On a platforms where signed long
is only 32 bits the address which is stored in temp_s will experience
overflow.


paulstelian97:
How are we sure values outside the range aren't possible in the input? If they are possible, you're hiding a bug.

PR #61, opened 2021-08-31
config: Adjust return type with function name

ktrzcinx:
Function with `parse_uint32` should return uint32_t value

Signed-off-by: Karol Trzcinski <karolx.trzcinski@intel.com>

### DIFF
--- a/src/adsp_config.c
+++ b/src/adsp_config.c
@@ -134,6 +134,7 @@ static uint32_t parse_uint32_hex_key(const toml_table_t *table, struct parse_ctx
 	if (!raw) {
 		if (def < 0 || def > UINT32_MAX) {
 			*error = err_key_not_found(key);
+			return UINT32_MAX;
 		} else {
 			*error = 0;
 			return (uint32_t)def;
@@ -181,6 +182,7 @@ static uint32_t parse_uint32_key(const toml_table_t *table, struct parse_ctx *ct
 	if (!raw) {
 		if (def < 0 || def > UINT32_MAX) {
 			*error = err_key_not_found(key);
+			return UINT32_MAX;
 		} else {
 			*error = 0;
 			return (uint32_t)def;
--- a/src/adsp_config.c
+++ b/src/adsp_config.c
@@ -169,8 +169,8 @@ static uint32_t parse_uint32_hex_key(const toml_table_t *table, struct parse_ctx
  * @param error code, 0 when success
  * @return default, parsed, or UINT32_MAX value
  */
-static int parse_uint32_key(const toml_table_t *table, struct parse_ctx *ctx, const char *key,
-			    int64_t def, int *error)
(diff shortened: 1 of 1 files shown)


paulstelian97 (approved):
I guess the function always succeeds??

ktrzcinx:
> I guess the function always succeeds??

No, it can fail. For checking this, there is `error` parameter.

paulstelian97:
A bit dubious but I guess we shouldn't make more important changes (I'm more used to having errors directly returned and results returned via pointer, rather than the converse, when both cannot be returned in a single value)

ktrzcinx:
>I'm more used to having errors directly returned and results returned via pointer

Me too, here the issue was a direct writing to different data types in structs, like uint8_t, uint16_t, or some enums without a risk of unintended modifying following fields in case of fields size mismatch. In solution like this, the result may be easily casted to different data type.

Moreover, such a way of returning error code, force developer to define `error` variable - otherwise code just won't compile. In the result it's less likely to just forget about error checking.

src/adsp_config.c
```diff
@@ -169,8 +169,8 @@ static uint32_t parse_uint32_hex_key(const toml_table_t *table, struct parse_ctx
  * @param error code, 0 when success
  * @return default, parsed, or UINT32_MAX value
  */
-static int parse_uint32_key(const toml_table_t *table, struct parse_ctx *ctx, const char *key,
-			    int64_t def, int *error)
+static uint32_t parse_uint32_key(const toml_table_t *table, struct parse_ctx *ctx, const char *key,
+				 int64_t def, int *error)
```
lgirdwood, line 173: Can we also update the doxygen to reflect the usage. i.e. on error a 0 is returned and ```error``` is set. Fwiw strtol has a similar calling convention except it sets ```errno```. TBH in parsing functions its always better to return the parsing status and have clients pass in an pointer for the parsing result (the unit32_t in this case)

ktrzcinx: >Can we also update the doxygen

Done

>strtol has a similar calling convention except it sets errno

In `strtol` API, there is hard to remember about error code on the contrary to this api, when it's hard to forget about it 😄 

> TBH in parsing functions its always better to return the parsing status and have clients pass in an pointer for the parsing result (the unit32_t in this case)

It's a common convention, but  I don't understand the reason why it is so.
Impossibility to create the memory corruption of following struct fields is a strong argument for this API, at least from my point of view.
Additional profit is less lines of source code, when error code should be bubbled up there is a field size mismatch.
See:

```
	foo->bazz_u8 = (uint8_t) parse_uint32_key(src, &status);
	if (status)
		return status;
```

vs

```
        uint32_t temp;
	status= parse_uint32_key(src, &temp); /* memory corruption when  foo->bazz_u8 is accessed directly */
                                              /* notice, that data type may not be a part of a field name */
	if (status)
		return status;
	foo->bazz_u8 = (uint8_t)temp;
```

lgirdwood (approved):

PR #67, opened 2021-09-14
ext_manifest: Fix incorrect signature

mrajwa:
This patch changes the incorrect signature of extended manifest
to the correct one "$AE1"

### DIFF
--- a/src/include/rimage/sof/kernel/ext_manifest.h
+++ b/src/include/rimage/sof/kernel/ext_manifest.h
@@ -33,8 +33,8 @@
 #define __packed __attribute__((packed))
 #endif
 
-/* In ASCII `XMan` */
-#define EXT_MAN_MAGIC_NUMBER	0x6e614d58
+/* In ASCII `$AE1` */
+#define EXT_MAN_MAGIC_NUMBER	0x31454124
 
 /* Build u32 number in format MMmmmppp */
 #define EXT_MAN_BUILD_VERSION(MAJOR, MINOR, PATH) ( \


lgirdwood (approved):
Good find ! Lets get this in v1.9.

mrajwa:
@kkarask, we would need to adjust tests for that as well. Can you handle it?

RanderWang:
@lgirdwood  @mrajwa  this EXT_MAN_MAGIC_NUMBER "XMan" is for SOF FW in ipc3.   please check https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/loader.c#L267 . This patch will make SOF ipc3 failed.  "AE1" is defined in cavs/cavs_ext_manifest.h, which is for IPC4.

mrajwa:
@RanderWang but where did you get "XMan" name from? This is invalid ID which causes SoF FW image analyze failure with tools like IIR for example.

RanderWang:
XMan is not designed by me. This is ID for current SOF.

RanderWang:
please check https://github.com/thesofproject/sof/blob/main/src/include/kernel/ext_manifest.h#L47

mrajwa:
Okay but this seems to be some out of blue name which as I said is not recognized by Intel Image Analyzer even - hence this analyzer says "sof.ri is corrupted". Wouldn't it be better to adjust driver to use the correct name asap?

RanderWang:
not only driver, but also fw.  @lgirdwood  we will go on support ipc3 and XMan is for it. And the hardest thing is that: this change will break the compatibility of released FW and people will complain that the release FW can't work with updated Linux OS. 
Last time I provided a PR and merged, but it broke the compatibility and at last the PR was reverted by @kv2019i.

RanderWang:
> Okay but this seems to be some out of blue name which as I said is not recognized by Intel Image Analyzer even - hence this analyzer says "sof.ri is corrupted". Wouldn't it be better to adjust driver to use the correct name asap?

do you use special config to build sof fw for windows driver ? The normal build is for Linux driver.

mrajwa:
Yes I have prepared dedicated config for WD but what I have said about the image analyzer is common for any sof build as of today.

lgirdwood:
Hmm, ok so we are using a different ID tag for manifests. We may ned to revert unless, we can make this a cmd line option for rimage (uses old ID by default unless target build config sets the ID option)

@mrajwa @RanderWang I assume the container structures are the same and the container IDs are the same ?

@ujfalusi looks the the upstream driver needs to support both IDs for the same object. Can you take care of this once @mrajwa and @RanderWang are aligned on the containers.

plbossart:
> @ujfalusi looks the the upstream driver needs to support both IDs for the same object. 

All existing firmware has to remain the same, and support for a new tag can only be added in the kernel for new platforms not yet upstream. Are we aligned?

ranj063:
> All existing firmware has to remain the same, and support for a new tag can only be added in the kernel for new platforms not yet upstream. Are we aligned?

but what about supporting the cavs FW on existing platforms @plbossart ? Don't we need this for that?

plbossart:
> > All existing firmware has to remain the same, and support for a new tag can only be added in the kernel for new platforms not yet upstream. Are we aligned?
> 
> but what about supporting the cavs FW on existing platforms @plbossart ? Don't we need this for that?

This is a development task, there are **no plans** to support the cAVS FW on GLK..ADL. Let's not confuse enablement and upstream releases, shall we?

RanderWang:
> > All existing firmware has to remain the same, and support for a new tag can only be added in the kernel for new platforms not yet upstream. Are we aligned?
> 
> but what about supporting the cavs FW on existing platforms @plbossart ? Don't we need this for that?

we support cavs FW in our cavs image parser, but we don't need to change current sof fw parser.

RanderWang:
> Yes I have prepared dedicated config for WD but what I have said about the image analyzer is common for any sof build as of today.
> 
> We can mark this change for feature release then.

can you share the image analyzer ?  I tried two, one is sof_ri_info and one is from cavs_fw

RanderWang:
for CAVS FW, we should go this function, and we will use EXTENDED_MANIFEST_MAGIC_HEADER_ID ($AE1), not "XMan".
````c
int ext_man_write_cavs_25(struct image *image)
{
	struct fw_image_ext_module *mod_ext;
	struct fw_ext_man_cavs_header header;

....

	header.version_major = mod_ext->ext_mod_config_array->header.version_major;
	header.version_minor = mod_ext->ext_mod_config_array->header.version_minor;
	header.num_module_entries = count;
	header.id = EXTENDED_MANIFEST_MAGIC_HEADER_ID;  // $AE1
	header.len = sizeof(const struct fw_ext_man_cavs_header);
````

PR #68, opened 2021-09-14
remove '-x' option

mrajwa:
This patch removes the unnecessary and unscalable -x option.
This option aims to add an offset to the .data section of
its parent .module section. This is horrible idea because this
offset will vary depend on the compiler used. It is better and
much easier to just store only the .data section so the offset
is always 0.

SoF FW related change -> https://github.com/thesofproject/sof/pull/4762


lgirdwood (approved):
Lets update the rimage submodule so we can get this and the header fix in v1.9

PR #70, opened 2021-09-17
Do not exit(0) on invalid options or parameters

marc-hb:
When some automation script gets some rimage flag wrong then rimage
exits and does not run at all. In such a case we want the build to fail
immediately and see the error at the bottom of the build logs and not
obscured by other, consecutive failures caused by the lack of rimage
output.

Take control of the exit code away from usage() and give it back to
main() where it belongs.

### DIFF
--- a/src/rimage.c
+++ b/src/rimage.c
@@ -30,7 +30,6 @@ static void usage(char *name)
 	fprintf(stdout, "\t -b build version\n");
 	fprintf(stdout, "\t -e build extended manifest\n");
 	fprintf(stdout, "\t -y verify signed file\n");
-	exit(0);
 }
 
 int main(int argc, char *argv[])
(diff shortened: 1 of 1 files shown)


marc-hb:
Tested in https://github.com/thesofproject/sof/pull/4778

aiChaoSONG:
basically, I think the usage function itself should not handle exit, the caller should do, in `-h`,  print usage and exit success, and other condition, print usage and exit failure. but this is another topic.

marc-hb:
Totally agree with you @aiChaoSONG but on the other hand this PR is changing a single character :-)

ktrzcinx (approved):

keyonjie:
@marc-hb I agree with @aiChaoSONG here, I prefer to remove the "exit(0)" from the usage(), and add 'exit(0)' to the '-h' 'exit(1)' to other callers.

marc-hb:
> I prefer to remove the "exit(0)" from the usage(), and add 'exit(0)' to the '-h' 'exit(1)' to other callers.

Done!

keyonjie (approved):
Looks good to me, thanks @marc-hb

iuliana-prodan (approved):

aiChaoSONG (approved):

lgirdwood (approved):

paulstelian97 (approved):

PR #73, opened 2021-09-21
README.md: add warning that forks must be created using Github

marc-hb:
Otherwise the SHA1 is not found and the rest of the process fails.

Problem discovered in
https://github.com/thesofproject/sof/pull/4790#discussion_r713326709

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,9 @@ to generate binary image files.
 
 ## Building
 
-The `rimage` tool can be built with the usual CMake commands:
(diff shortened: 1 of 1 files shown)


cujomalainey (approved):
makes sense, being able to run arbitrary on commits on submodules would be bad.

mrajwa (approved):

dbaluta (approved):

lgirdwood (approved):

PR #76, opened 2021-11-16
[RFC] rimage: add functionality to resign fw image

juimonen:
Add function to resign already signed images. This can
be done with switch "q" as follows:

"rimage -q sof-tgl.ri -o sof-tgl-mod.ri -c tgl.toml -k key.pem"

Signed-off-by: Jaska Uimonen <jaska.uimonen@linux.intel.com>

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -1427,3 +1427,127 @@ out:
 	fclose(in_file);
 	return 0;
 }
+
+
+int resign_image(struct image *image)
+{
+	int key_size, key_file_size;
+	size_t size, read;
+	FILE *in_file;
+	void *buffer;
+	int ret, i;
+
+	/* open image for reading */
+	in_file = fopen(image->in_file, "rb");
+	if (!in_file) {
+		fprintf(stderr, "error: unable to open %s for reading %d\n",
+			image->in_file, errno);
+		return -errno;
+	}
+
+	/* get file size */
+	ret = fseek(in_file, 0, SEEK_END);
+	if (ret < 0) {
+		fprintf(stderr, "error: unable to seek eof %s for reading %d\n",
+			image->verify_file, errno);
+		ret = -errno;
+		goto out;
+	}
+
+	size = ftell(in_file);
+	if (size < 0) {
+		fprintf(stderr, "error: unable to get file size for %s %d\n",
+			image->verify_file, errno);
+		ret = -errno;
+		goto out;
+	}
+
+	ret = fseek(in_file, 0, SEEK_SET);
+	if (ret < 0) {
+		fprintf(stderr, "error: unable to seek %s for reading %d\n",
+			image->verify_file, errno);
+		ret = -errno;
+		goto out;
+	}
+
+	/* allocate buffer for parsing */
+	buffer = malloc(size);
+	if (!buffer) {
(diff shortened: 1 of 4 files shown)


lgirdwood (approved):

lgirdwood:
Will wait as discussed for the manifest check to be included in PR

lgirdwood (approved):

marc-hb:
@juimonen  it looks like this PR added a lot of clang warnings (found by mere chance)

```
sof/rimage/src/manifest.c:1467:6: error: variable 'buffer' is used uninitialized whenever 'if' condition is true
      [-Werror,-Wsometimes-uninitialized]
        if (ret < 0) {
            ^~~~~~~
sof/rimage/src/manifest.c:1552:7: note: uninitialized use occurs here
        free(buffer);
             ^~~~~~
sof/rimage/src/manifest.c:1467:2: note: remove the 'if' if its condition is always false
        if (ret < 0) {
        ^~~~~~~~~~~~~~
sof/rimage/src/manifest.c:1459:6: error: variable 'buffer' is used uninitialized whenever 'if' condition is true
      [-Werror,-Wsometimes-uninitialized]
        if (size < 0) {
            ^~~~~~~~
sof/rimage/src/manifest.c:1552:7: note: uninitialized use occurs here
        free(buffer);
             ^~~~~~
sof/rimage/src/manifest.c:1459:2: note: remove the 'if' if its condition is always false
        if (size < 0) {
        ^~~~~~~~~~~~~~~
sof/rimage/src/manifest.c:1451:6: error: variable 'buffer' is used uninitialized whenever 'if' condition is true
      [-Werror,-Wsometimes-uninitialized]
        if (ret < 0) {
            ^~~~~~~
sof/rimage/src/manifest.c:1552:7: note: uninitialized use occurs here
        free(buffer);
             ^~~~~~
sof/rimage/src/manifest.c:1451:2: note: remove the 'if' if its condition is always false
        if (ret < 0) {
        ^~~~~~~~~~~~~~
sof/rimage/src/manifest.c:1438:14: note: initialize the variable 'buffer' to silence this warning
        void *buffer;
                    ^
                     = NULL
```

src/manifest.c
```diff
@@ -1427,3 +1427,127 @@ int verify_image(struct image *image)
...
 }
+
+
+int resign_image(struct image *image)
+{
+	int key_size, key_file_size;
+	size_t size, read;
+	FILE *in_file;
+	void *buffer;
+	int ret, i;
```
marc-hb, line 1438: rimage does not live in the previous century like the kernel does, so you can move this at first point of use and combine declaration and initialization as allowed by C99

juimonen: I guess all this goes away if just setting this to NULL at declaration as clang suggest... I mean it is a clear bug to free some bogus

juimonen: @marc-hb made this: https://github.com/thesofproject/rimage/pull/78.

marc-hb: Initializing all variables with some default value at the top makes sense... when and only when some default value actually makes sense. In other cases it merely hides compiler warnings finding bugs for C99 backward-compatibility (I didn't read that code in that PR).

PR #78, opened 2021-12-01
manifest: fix buffer init in resign

juimonen:
Set buffer to NULL to prevent freeing bogus memory in error case.

Signed-off-by: Jaska Uimonen <jaska.uimonen@linux.intel.com>


marc-hb (approved):
No more warnings with clang 12

lgirdwood:
> No more warnings with clang 12

Looks like we need to add clang to CI ?

lgirdwood (approved):

PR #87, opened 2022-01-25
Resurrect Windows build and upgrade CMake version to match Zephyr

aborisovich:
Two commits:
1) Added handling of OpenSSL on Windows using OPENSSL_PATH.
Upgraded CMake minimal version to 3.20.0 to match Zephyr and to use cmake_path freely.
2) Fixed warning treated as error in src/manifest.c where variable passed to fprintf was of type size_t and expected format of long int.

Signed-off-by: Andrey Borisovich [andrey.borisovich@intel.com](mailto:andrey.borisovich@intel.com)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,27 @@ if(CMAKE_C_COMPILER_VERSION VERSION_GREATER 9.1)
 	target_compile_options(rimage PRIVATE -Wno-char-subscripts)
 endif()
 
+# Windows builds use MSYS2 https://www.msys2.org/ to get linux tools and headers.
+# MSYS_INSTALL_DIR variable points to MSYS2 installation directory.
+# You may pass it as environmental or cmake configure variable.
+if(${CMAKE_HOST_WIN32})
+	cmake_minimum_required(VERSION 3.20)
+	if(DEFINED ENV{MSYS_INSTALL_DIR} AND NOT MSYS_INSTALL_DIR)
+		set(MSYS_INSTALL_DIR $ENV{MSYS_INSTALL_DIR})
+	endif()
+
+	if(MSYS_INSTALL_DIR)
+		cmake_path(IS_ABSOLUTE MSYS_INSTALL_DIR IS_MSYS_INSTALL_DIR_ABSOLUTE)
+		if(NOT IS_MSYS_INSTALL_DIR_ABSOLUTE)
+			message(FATAL_ERROR "Please provide absolute path to MSYS2 installation
+				setting MSYS_INSTALL_DIR env variable")
+		endif()
+		# Include standard posix headers. Requires pacman openssl-devel package.
+		cmake_path(APPEND MSYS_INSTALL_DIR "usr" "include" OUTPUT_VARIABLE MSYS_SYSTEM_INCLUDE_PATH)
+		target_include_directories(rimage PRIVATE "${MSYS_SYSTEM_INCLUDE_PATH}")
+	endif()
+endif()
+
 target_link_libraries(rimage PRIVATE crypto)
 
 target_include_directories(rimage PRIVATE
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
-// Copyright(c) 2018 Intel Corporation. All rights reserved.
+// Copyright(c) 2018-2022 Intel Corporation. All rights reserved.
 //
 // Author: Liam Girdwood <liam.r.girdwood@linux.intel.com>
 //         Keyon Jie <yang.jie@linux.intel.com>
@@ -1480,7 +1480,7 @@ int resign_image(struct image *image)
 	/* read file into buffer */
 	read = fread(buffer, 1, size, in_file);
 	if (read != size) {
-		fprintf(stderr, "error: unable to read %ld bytes from %s err %d\n",
+		fprintf(stderr, "error: unable to read %zu bytes from %s err %d\n",
 					size, image->in_file, errno);
 		ret = errno;
 		goto out;


src/manifest.c
```diff
@@ -1480,7 +1480,7 @@ int resign_image(struct image *image)
 	/* read file into buffer */
 	read = fread(buffer, 1, size, in_file);
 	if (read != size) {
-		fprintf(stderr, "error: unable to read %ld bytes from %s err %d\n",
+		fprintf(stderr, "error: unable to read %zu bytes from %s err %d\n",
```
marc-hb, line 1483: This belongs to a different commit.

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.20.0)
```
marc-hb, line 3: This is probably going to cause some CI failures.

marc-hb (commented):
Please submit this to SOF CIs, see rimage/README (and recent example in #88)

CMakeLists.txt
aborisovich: I had submitted a request to SOF infrastructure team for cmake upgrade.

lgirdwood: @marc-hb @greg-intel I assume Jenkins is good for this cmake version ?

lgirdwood (commented):
Pls add your SoB.

CMakeLists.txt
marc-hb: It seems to be failing not just in Jenkins: https://github.com/thesofproject/sof/pull/5265

marc-hb: Can you require 3.20 for `CMAKE_HOST_WIN32` only? According to your commit message it does not seem needed for Linux.

lgirdwood: @marc-hb we are moving to Zephyr being the default rtos. We need to plan and upgrade the containers so that all developers can build for Zephyr. 
Let also assume that container updates will be ad-hoc and follow upstream development. This should be factored into our processes so it's a painless as possible.

marc-hb: We offer only one crosstool-ng container for developers, so this single one is not the problem. The problem is the large variety of Linux build environments (not just containers) across XTOS developers and XTOS build machines in multiple CIs maintained by different owners, some we don't even know about. Breaking the build for _all_ these legacy XTOS environments  unless they upgrade to cutting edge CMake version is in aggregate a huge disruption and amount of work for... just one Windows-only command? Really not worth it when this command can be guarded and used by Windows only. Seems like a no-brainer.

Zephyr requested a cutting edge CMake version because it really needed some critical CMake features yet users were not happy. Unlike XTOS, Zephyr developers got at least a very active stream of new features in return.

Off-topic but note many developers do not use our crosstool-ng container at all, because either
- they use xcc, or
- they built crosstool-ng themselves to avoid the docker layer of indirection and immutability of the container (very painful to install extra packages or any other image change), or
- they use Zephyr which does not compile with our crosstool-ng.

BTW the Zephyr SDK is an open-source project also based on crosstool-ng, so why are still maintaining our own crosstool-ng fork? (and container?) Or should we at least rebase our crosstool-ng on top of the Zephyr SDK? Off-topic sorry.

greg-intel: > @marc-hb @greg-intel I assume Jenkins is good for this cmake version ?

Five of the build servers, sofbld07.ostc.intel.com through sofbld11.ostc.intel.com are running cmake version 3.22.1, but the others are using 3.16.3, so it's hit or miss.

aborisovich: > Breaking the build for all these legacy XTOS environments unless they upgrade to cutting edge CMake version is in aggregate a huge disruption and amount of work for... just one Windows-only command? Really not worth it when this command can be guarded and used by Windows only. Seems like a no-brainer.

Windows command is not a cause of this change but a reason to do it. So is Zephyr. I believe that this is what software maintenance is about - we constantly try to update all our tools not to get obsolete. Same with python, testing frameworks etc. It is good to perform this step now before further build system refactorings so we could check all our CI supported configurations are passing after the upgrade. Besides as @greg-intel mentioned, some servers already run CMake 3.22.1.

plbossart:
@aborisovich keep in mind that not all developers chase the latest versions of everything, be it gcc, cmake, alsa-lib, alsa-utils or whatnot.

I for one have

cmake --version
cmake version 3.16.3

please minimize disruption for *EXISTING* environments, thank you.

CMakeLists.txt
marc-hb: > So is Zephyr.

Zephyr is totally off-topic here as it already requires 3.21 (which didn't please a number of users). This particular discussion is about XTOS developers and XTOS only.

> Windows command is not a cause of this change but a reason to do it. 

So move cmake_minimum_required(3.20) to the `if(CMAKE_HOST_WIN32)` section for now and revisit and annoy ALL Linux users (giving Intel a bad name) later when you have much stronger reason(s) to do so. If you wait long enough then 3.20 will be available in most Linux distributions by default and this will have become a non-issue. How inconvenient is this to you and how does that inconvenience compares to the inconvenience of every single XTOS developer, virtual machine and CI system inside and outside Intel to add one extra repository and upgrade one of their most fundamental development packages, potentially affecting all their other development efforts?

> I believe that this is what software maintenance is about - we constantly try to update all our tools not to get obsolete. 

This is clearly the point of view of a developer: someone who enjoys writing and rewriting code and can afford to spend most of his time doing that. People swamped with applying security fixes and finding and reporting regressions have a very different perspective.

This vision of software maintenance rings truer to me:
https://medium.com/@sdboyer/so-you-want-to-write-a-package-manager-4ae9c17d9527

> Despite the lack of necessary relationship between versions and code, there is at least one way in which versions help quite directly to ensure software works: they focus public attention on a few particular revisions of code. With attention focused thusly, Linus’ Law suggests that bugs will be rooted out and fixed (and released in subsequent patch versions). In this way, the curatorial effect of focusing attention on particular versions reduces systemic risk around those versions. This helps with another of FLOSS’ uncertainties:

> > I have to prepare myself for the likelihood that most of what’s out there will probably be crap. Sorting wheat from chaff will also take time.

> Having versions at least ensures that, if the software is crap, it’s because it’s actually crap, not because you grabbed a random revision that happened to be crap. **That saves time. Saved time can save projects.**

aborisovich: This discussion is going nowhere.
Thankfully we do not have to agree and you are not decisive person here.
> So move cmake_minimum_required(3.20) to the if(CMAKE_HOST_WIN32) section for now and revisit and annoy ALL Linux users (giving Intel a bad name) later when you have much stronger reason(s) to do so.

Look at Intel slogans. I see there "Fearless". I see there "newest technologies".
Do you think your approach towards cmake aligns with this policy and culture?
We need to go forward not lag behind. It's not our problem that linux package managers lags behind.
What kind of approach it is? You install software - always newest stable version. Thats trivial and obvious for everyone - for Windows users, for Mac users, for Android users but no, not for linux package managers.

aborisovich:
Lets not confuse disruption with maintenance please.
Especially that no regression had been found as of cmake update concern.
@plbossart now you have a reason to upgrade!

aborisovich:
Yeah ok I am tired arguing with you.
I will put that under if windows.

plbossart:
> Yeah ok I am tired arguing with you. I will put that under if windows.

thank you for listening to feedback @aborisovich, much appreciated.

CMakeLists.txt
marc-hb: > Do you think your approach [...] ? It's not our problem that linux package managers lags behind.

I'm absolutely not claiming Linux package management is the best approach[*], please stop identify me with that. Don't confuse me with Ubuntu, I'm just the messenger. I'm merely describing where all existing SOF users are, inside and outside Intel and I think they matter. In fact many Linux users are not happy either: https://lwn.net/Articles/867657/

Package management is always very messy no matter how you do it and Windows is no exception. One of its (universal) aspect is called "DLL hell". It's not a solved problem and will never be. There are pros and cons in all approaches. I highly recommend Sam Boyer's article I referenced above if you're interested in a broader perspective and more informed and nuanced opinions. 

[*] there are "rolling" distros too like Arch, Gentoo and others but they are much less popular.

> Thats obvious for Windows users,...

Microsoft is very actively supporting WSL2 and the default choice is Ubuntu; I find that very interesting. BTW does Visual Studio finally support C99?
> Look at Intel slogans. I see there "Fearless". I see there "newest technologies".

Intel also has a generally good standing in open-source communities outside Intel and "Wintel"'s backward compatibility has no equivalent in the entire industry.

This topic is one of the most complicated.

lgirdwood (approved):

lgirdwood:
@marc-hb good for you now ?

plbossart (changes requested):
@aborisovich you need to add a Signed-off-by tag for all open-source contributions.

CMakeLists.txt
```diff
@@ -34,6 +39,43 @@ if(CMAKE_C_COMPILER_VERSION VERSION_GREATER 9.1)
...
+		"${OPENSSL_BIN_PATH}/libcrypto-1_1-x64.dll")
+	set_property(TARGET crypto PROPERTY IMPORTED_IMPLIB_DEBUG
+		"${OPENSSL_LIB_PATH}/libcrypto.lib")
+	set_property(TARGET crypto PROPERTY IMPORTED_LOCATION_DEBUG
+		"${OPENSSL_BIN_PATH}/libcrypto-1_1-x64.dll")
+
+	target_sources(rimage PRIVATE "${OPENSSL_INCLUDE_PATH}/openssl/applink.c")
+	set_source_files_properties("${OPENSSL_INCLUDE_PATH}/openssl/applink.c" PROPERTIES
+		COMPILE_OPTIONS "-Wno-missing-prototypes")
+	target_compile_options(rimage PRIVATE -Wl,-EL)
```
marc-hb, line 72: Can you explain why `-EL` is needed? rimage runs on the system that builds SOF so I wouldn't expect any endianness issue there. Keep in mind many people are not familiar with Windows around here (in case that wasn't obvious yet...)

I'm also curious what toolchain you tested this with.

CMakeLists.txt
```diff
@@ -34,6 +39,43 @@ if(CMAKE_C_COMPILER_VERSION VERSION_GREATER 9.1)
...
+
+	add_custom_command (TARGET rimage POST_BUILD
```
marc-hb, line 74: That looks like an "installation" command, is it? Either way could you add a comment?

CMakeLists.txt
```diff
@@ -34,6 +39,43 @@ if(CMAKE_C_COMPILER_VERSION VERSION_GREATER 9.1)
...
+		COMMAND ${CMAKE_COMMAND} -E copy_if_different
+		$<TARGET_FILE:crypto> $<TARGET_FILE_DIR:rimage>)
```
marc-hb, line 76: Is that enough to generate the rimage -> crypto build dependency? Apologies for the naive CMake question, you probably have more CMake knowledge than all other persons currently contributing to rimage combined.

CMakeLists.txt
aborisovich: Yes that copied dll library to install dir. Not applicable anymore - i took new simpler approach with MSYS2 shell and openssl is taken from there.

CMakeLists.txt
aborisovich: Yes we had mismatches as Windows is little-endian and we gcc we used on Windows was form MSYS2.
We use https://www.msys2.org/ which allows mixing posix tools with Windows easily.
Eg. you install MSYS2 to C:\msys64 and this separate shell establishes linux root dir there.
But they fun part is that executable binaries are windows .exe format.
So you end up with adding to Windows PATH C:\msys64\usr\bin and can easily invoke posix tools like gcc.exe.

marc-hb:
> @marc-hb good for you now ?

Yes especially with the new and much shorter MSYS approach. Not formally approving because this entire PR is really something for Windows developer(s) to review and I'm not one at all.

marc-hb:
> Upgraded CMake minimal version to 3.20.0 to match Zephyr and to use cmake_path freely.

Just noticed by chance that `cmake_path` is buggy in some CMake versions: https://github.com/zephyrproject-rtos/zephyr/commit/8b199b0fbb253aec897

aborisovich:
> > Upgraded CMake minimal version to 3.20.0 to match Zephyr and to use cmake_path freely.
> 
> Just noticed by chance that `cmake_path` is buggy in some CMake versions: [zephyrproject-rtos/zephyr@8b199b0fbb253aec897](https://github.com/zephyrproject-rtos/zephyr/commit/8b199b0fbb253aec897)

In some CMake *distributions* - exactly PIP distributions here. You can just install from https://cmake.org/download/ using .sh script and manually add to PATH. Well, what can I say... "You never know who is going to break your software" :-(

Issue #86, opened 2022-01-25
intel_adsp_cavs25: output some unrecognized ASCII char when using rimage to sign a binary


chen-png:
**Describe the bug**

for intel_adsp_cavs25 platform, when using rimage to sign the binary, it reported some unrecognized ASCII char when parsing cse entry name "cavs0015.met", the output as below, there are some unrecognized char after "cavs0015.met".
please ignore data of entry.offset and length, I removed them due to private policy.

```
 partition_name: 'ADSP'
         header_version: 2
          entry_version: 1
             nb_entries: 3
             entry.name: 'ADSP.man'
           entry.offset: 
           entry.length:
             entry.name: 'cavs0015.met▒'
           entry.offset: 
           entry.length: 
             entry.name: 'cavs0015'
           entry.offset: 
           entry.length: 
```

this will causing an error "UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc0 in position 899: invalid start byte" when we want to decode the output of rimage command.
the raw data captured at position 899 as below, cannot decode \xc0\x04 using utf-8.

```
entry.name: 'cavs0015.met\xc0\x04'\n
```

**Debug Info**
I found that in tgl.toml file, there is a cse entry as below, the strlen(name) is 12 bytes,
```
[[cse.entry]]
name = "cavs0015.met"
offset = 
length = 
```
but the **cse entry definition in rimage source code is as below, the entry name is uint8 [12], it includes the empty char for string**. so when rimage parsing entry name "cavs0015.met", it just overlay the empty char, then when we output this entry_name, it will output some other chars until meeting another empty char.
so when we define a cse entry name, the max strlen(name) should be 11 chars, not 12 chars.
```
struct CsePartitionDirEntry {
        uint8_t  entry_name[12];
        uint32_t offset;
        uint32_t length;
        uint32_t reserved;
}  __attribute__((packed));
```
**Suggestions**
**Could we increase the bytes of cse entry name in rimage source code?**

**Another minimal issue**
this seems like a mistype.
```
diff --git a/src/adsp_config.c b/src/adsp_config.c
index 2e6f39f..5d5aa5a 100644
--- a/src/adsp_config.c
+++ b/src/adsp_config.c
@@ -627,7 +627,7 @@ static int parse_cse_v2_5(const toml_table_t *toml, struct parse_ctx *pctx,

        hdr->nb_entries = toml_array_nelem(cse_entry_array);

-       if (1 || verbose)
+       if (verbose)
[...]

PR #88, opened 2022-01-26
Fix confusion in headers between uint8_t arrays and strings

marc-hb:
2 commits. Main one:

Fix confusion in headers between uint8_t arrays and strings 

Strings are null terminated, byte arrays not always.

Fixes garbage in debug output issue #86

Change parse_str_key() argument from char * to uint8_t *. Casting the
argument on almost every call was a clear indication of the problem.

Add new DUMP_PRINTABLE_BYTES() to printf uint8_t arrays _and_ their
_optional_ padding.

No change to the .ri output, bit for bit identical.

Here's an example of how the -v output is fixed:

```diff
    mem_zone.host_offset: 0x0

 cse
-         partition_name: 'ADSPADSP.man'
+         partition_name: ADSP
          header_version: 1
           entry_version: 1
              nb_entries: 3
-             entry.name: 'ADSP.man'
+             entry.name: ADSP.man\x00\x00\x00\x00
            entry.offset: 0x58
            entry.length: 0x378
-             entry.name: 'cavs0015.met'
+             entry.name: cavs0015.met
            entry.offset: 0x400
            entry.length: 0x60
-             entry.name: 'cavs0015'
+             entry.name: cavs0015\x00\x00\x00\x00
            entry.offset: 0x480
            entry.length: 0x0

@@ -140,7 +132,7 @@
           exponent_size: 1

 signed_pkg
-                   name: 'ADSP'
+                   name: ADSP
                     vcn: 0
                     svn: 0
                 fw_type: 0
@@ -161,17 +153,17 @@
                  bitmap: 0
                  bitmap: 0
                  bitmap: 0
-              meta.name: 'cavs0015.met^C^B '
+              meta.name: cavs0015.met
               meta.type: 0x3
          meta.hash_algo: 0x2
          meta.hash_size: 0x20
          meta.meta_size: 96

 partition_info
-                   name: 'ADSP'
+                   name: ADSP
            part_version: 0x10000000
             instance_id: 1
-            module.name: 'cavs0015.met^C'
+            module.name: cavs0015.met
        module.meta_size: 0x60
             module.type: 0x3

@@ -186,7 +178,7 @@

 fw_desc.header
               header_id: '$AM1'
-                   name: 'ADSPFW'
+                   name: ADSPFW\x00\x00
      preload_page_count: 0
          fw_image_flags: 0x0
            feature_mask: 0xffff
```

### DIFF
--- a/src/adsp_config.c
+++ b/src/adsp_config.c
@@ -20,11 +20,33 @@
 #include <string.h>
 #include <stdio.h>
 #include <errno.h>
+#include <ctype.h>
 
 /* macros used to dump values after parsing */
 #define DUMP_KEY_FMT "   %20s: "
 #define DUMP(fmt, ...) fprintf(stdout, fmt "\n", ##__VA_ARGS__)
 #define DUMP_KEY(key, fmt, ...) DUMP(DUMP_KEY_FMT fmt, key, ##__VA_ARGS__)
+
+static void print_bytes(FILE *out, const uint8_t *arr, size_t len)
+{
+	for (const uint8_t *pos = arr; pos < arr + len; pos++) {
+		char c = *pos;
+
+		if (isprint(c))
+			fputc(c, out);
+		else
+			fprintf(out, "\\x%.2x", c);
+	}
+}
+
+#define DUMP_PRINTABLE_BYTES(name, var) _dump_printable_bytes(name, var, sizeof(var))
+
+static void _dump_printable_bytes(const char *name, const uint8_t *arr, size_t len)
+{
+	printf(DUMP_KEY_FMT, name);
+	print_bytes(stdout, arr, len);
+	printf("\n");
+}
 
 /** parser counter, used to assert nothing left unparsed in toml data */
 struct parse_ctx {
@@ -206,16 +228,20 @@ static uint32_t parse_uint32_key(const toml_table_t *table, struct parse_ctx *ct
 }
 
 /**
- * Parse string value from key in given toml table
+ * Parse string value from key in given toml table to uint8_t array. The
+ * destination is NOT a string because it is padded with zeros if and
+ * only if there is some capacity left. For string destinations use
+ * parse_str_key().
+ *
  * @param table toml table where key is specified
  * @param ctx parsing context, key counter will be incremented after successful key parse
  * @param key field name
- * @param dst target buffer for string
- * @param capacity dst buffer size
+ * @param dst uint8_t[] destination
+ * @param capacity dst array size
  * @param error code, 0 when success
  */
-static void parse_str_key(const toml_table_t *table, struct parse_ctx *ctx, const char *key,
-			  char *dst, int capacity, int *error)
+static void parse_printable_key(const toml_table_t *table, struct parse_ctx *ctx, const char *key,
+				uint8_t *dst, int capacity, int *error)
 {
 	toml_raw_t raw;
 	char *temp_s;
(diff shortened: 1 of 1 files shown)


andyross (approved):
Seems very reasonable.

chen-png (approved):
LGTM, also tested and it fixed #86.

RanderWang (approved):

src/adsp_config.c
```diff
@@ -26,6 +26,27 @@
 #define DUMP(fmt, ...) fprintf(stdout, fmt "\n", ##__VA_ARGS__)
 #define DUMP_KEY(key, fmt, ...) DUMP(DUMP_KEY_FMT fmt, key, ##__VA_ARGS__)
 
+static void print_bytes(FILE *out, const uint8_t *arr, size_t len)
+{
+	for (const uint8_t *pos = arr; pos < arr + len; pos++) {
+		char c = *pos;
+
+		if (' ' <= c && c <= '~')
```
ktrzcinx, line 34: maybe `isprint` ?

src/adsp_config.c
```diff
@@ -374,7 +399,9 @@ static int parse_adsp(const toml_table_t *toml, struct parse_ctx *pctx, struct a
 		/* non-configurable fields */
 
 		/* configurable fields */
-		parse_str_key(mem_zone, &ctx, "type", zone_name, sizeof(zone_name), &ret);
+		zone_name[sizeof(zone_name) - 1] = 0;
+		parse_str_key(mem_zone, &ctx, "type", (uint8_t *)zone_name,
+			      sizeof(zone_name) - 1, &ret);
```
ktrzcinx, line 404: it seems reasonable to move clearing last string-array byte to `parse_str_key()` function... or even `memset` to zero what will guarantee deterministic output when such a mechanism is not implemented on enclosing structures.

src/adsp_config.c
```diff
@@ -1971,12 +1998,14 @@ static int parse_module(const toml_table_t *toml, struct parse_ctx *pctx,
...
-			      SOF_MAN_MOD_NAME_LEN, &ret);
+		parse_str_key(mod_entry, &ctx_entry, "name", mod_man->name,
+			      sizeof(mod_man->name), &ret);
 		if (ret < 0)
 			return err_key_parse("name", NULL);
 
-		parse_str_key(mod_entry, &ctx_entry, "uuid", buf, 48, &ret);
+		buf[sizeof(buf) - 1] = 0;
+		parse_str_key(mod_entry, &ctx_entry, "uuid", (uint8_t *)buf, sizeof(buf) - 1,
+			      &ret);
```
ktrzcinx, line 2008: ditto

src/adsp_config.c
ktrzcinx: > The destination is simply not a string.

It's not a string but it needs null termination here... so it's quite slippery. Setting null termination manually seems little ugly for me. Maybe a cleanest solution will be to just rename this parsing function to something like `parse_printable_key` (version without adding a null termination) and return length of a string then create a simple wrapper function:

```
static void parse_str_key(const toml_table_t *table, struct parse_ctx *ctx, const char *key,
			  char *dst, int capacity, int *error)
{
    int len = parse_printable_key(table, ctx, key, dst, capacity - 1, error);
    dst[len] = 0;
}
```

It's just and idea.

dcpleung (approved):

src/adsp_config.c
marc-hb: Thx! Done.

src/adsp_config.c
marc-hb: Done something close to what @ktrzcinx suggested please have a look.

Also added the too long string to the "too long" error message.

src/adsp_config.c
```diff
@@ -237,19 +263,50 @@ static void parse_str_key(const toml_table_t *table, struct parse_ctx *ctx, cons
...
-				 capacity);
+		if (len > 20) {
+			static const char ellipsis[] = "...";
+			const size_t el_len = sizeof(ellipsis);
+
+			strncpy(temp_s + len - el_len, ellipsis, el_len);
+		}
+
+		*error = log_err(-EINVAL, "Too long input '%s' for key '%s' (%d > %d) characters\n",
+				 temp_s, key, len, capacity);
```
ktrzcinx, line 274: I guess this part should truncate `temp_s` to 20 (17) chars and add `...` suffix.  Then it should be `strncpy(temp_s + max_len - el_len,....`

ktrzcinx (changes requested):

src/adsp_config.c
marc-hb: > I guess this part should truncate temp_s to 20 (17) chars and add ... suffix.

I don't understand, that's what the code above does.

What is `max_len`?

marc-hb: Fixed, thanks @ktrzcinx for the thorough review.

ktrzcinx (approved):
Now looks good, thanks !

PR #91, opened 2022-04-30
Write firmware file micro version to manifest for cAVS platforms

mengdonglin:
Fix https://github.com/thesofproject/rimage/issues/90

SOF CMake defines SOF_MAJOR, SOF_MINOR and SOF_MICRO for 3 fields
of firmware file version major.minor.micro

This patch changes hotfix_version to micro_version in headers to align
with SOF CMake, and adds fw_ver_micro to firmware image context.

If SOF CMake passes firmware file micro version in the version string,
rimage will write it to different manifest headers for cAVS platforms.
If micro version isn't passed to rimage, it's 0 by default.

This feature will allow sof_ri_info.py to dump entire file version from
a FW binary for cAVS platforms.

There will be a SOF PR to let CMake pass FW file micro version to rimage.

### DIFF
--- a/src/include/rimage/cavs/cavs_ext_manifest.h
+++ b/src/include/rimage/cavs/cavs_ext_manifest.h
@@ -230,6 +230,7 @@ struct fw_ext_man_cavs_header {
 	uint32_t len;	/* sizeof(Extend Manifest) in bytes */
 	uint16_t version_major;	/* Version of Extended Manifest structure */
 	uint16_t version_minor;	/* Version of Extended Manifest structure */
+	uint16_t version_micro; /* Version of Extended Manifest structure */
 	uint32_t num_module_entries;
 } __attribute__((packed));
 
@@ -238,7 +239,7 @@ struct fw_ext_mod_config_header {
 	uint32_t guid[4];	/* Module GUID */
 	uint16_t version_major;	/* Module version */
 	uint16_t version_minor;	/* Module version */
-	uint16_t version_hotfix;	/* Module version */
+	uint16_t version_micro;	/* Module version */
 	uint16_t version_build;	/* Module version */
 	enum mod_type module_type;
 	uint32_t init_settings_min_size;	/* Minimum size of initialization settings (in bytes) */
--- a/src/rimage.c
+++ b/src/rimage.c
@@ -115,19 +115,21 @@ int main(int argc, char *argv[])
 		return -EINVAL;
 	}
 
-	/* firmware version and build id */
+	/* firmware version that can be major.minor or major.minor.micro */
 	if (image.fw_ver_string) {
-		ret = sscanf(image.fw_ver_string, "%hu.%hu",
+		ret = sscanf(image.fw_ver_string, "%hu.%hu.%hu",
 			     &image.fw_ver_major,
-			     &image.fw_ver_minor);
+			     &image.fw_ver_minor,
+			     &image.fw_ver_micro);
 
-		if (ret != 2) {
+		if (ret != 2 && ret != 3) {
 			fprintf(stderr,
 				"error: cannot parse firmware version\n");
 			return -EINVAL;
 		}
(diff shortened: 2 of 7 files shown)


mengdonglin:
https://github.com/thesofproject/sof/pull/5760 is the PR let CMake pass FW file micro version to rimage.

src/include/rimage/cavs/cavs_ext_manifest.h
```diff
@@ -230,6 +230,7 @@ struct fw_ext_man_cavs_header {
 	uint32_t len;	/* sizeof(Extend Manifest) in bytes */
 	uint16_t version_major;	/* Version of Extended Manifest structure */
 	uint16_t version_minor;	/* Version of Extended Manifest structure */
+	uint16_t version_micro; /* Version of Extended Manifest structure */
```
lgirdwood, line 233: Wont this break manifest parsing on Windows ? @RanderWang can you comment.

src/rimage.c
```diff
@@ -115,19 +115,21 @@ int main(int argc, char *argv[])
...
 	if (image.fw_ver_string) {
-		ret = sscanf(image.fw_ver_string, "%hu.%hu",
+		ret = sscanf(image.fw_ver_string, "%hu.%hu.%hu",
 			     &image.fw_ver_major,
-			     &image.fw_ver_minor);
+			     &image.fw_ver_minor,
+			     &image.fw_ver_micro);
 
-		if (ret != 2) {
+		if (ret != 2 && ret != 3) {
```
aiChaoSONG, line 125: The return value of `sscanf` is the number of receiving arguments successfully assigned or EOF, I don't think we should allow partial assignment. the ret must be 3

mengdonglin: @aiChaoSONG Thanks for your comments. This is to be compatible with current SOF CMake that only passes 2 version arguments (major.minor) to rimage. If this rimage PR gets merged at first,  SOF building will fail as CMake doesn't pass 3 version arguments (major.minor.micro) as rimage expects. 

If SOF CMake https://github.com/thesofproject/sof/pull/5760 can also be merged, I can update rimage to only support 3 version arguments.

aiChaoSONG: Even the PR get merged, it will not be used, rimage is a submodule of SOF, unless rimage is upgraded on SOF side, new changes will never be used. so it's safe.

mengdonglin: I see. Thanks! I Will update PR after Rander comments if this PR will impact Windows.

src/include/rimage/cavs/cavs_ext_manifest.h
mengdonglin: @lgirdwood Rander is on vacation this week. Let's wait for his comments next week. Hope I'm not breaking Windows.

RanderWang: Windows driver will take version_micro as num_module_entries we don't change header file used by windows, and this will result to error.

aiChaoSONG: > Windows driver will take version_micro as num_module_entries we don't change header file used by windows, and this will result to error.

This only applies to the extended manifest. So I think we can skip the micro version only for cavs extended manifest.

mengdonglin: To avoid breaking Windows, I wrote a new PR #92 that will not change manifest header definitions but write the micro version to the unused hotfix_version field of manifest header. Can you help review the new PR#92?

src/rimage.c
mengdonglin: Fixed this in new PR#92 @aiChaoSONG

mengdonglin:
Close this PR and submitted a new PR #92 to address all review comments. @lgirdwood  @aiChaoSONG @RanderWang

PR #92, opened 2022-05-04
Write firmware file micro version to manifest for cAVS platforms

mengdonglin:
Fix https://github.com/thesofproject/rimage/issues/90
This is v2 of https://github.com/thesofproject/rimage/pull/91

SOF CMake can extract FW file version from latest git tag and defines
SOF_MAJOR, SOF_MINOR and SOF_MICRO for 3 version fields. But rimage only
gets major and minor version from SOF CMake and writes them into the
standard firmware manifest header of cAVS platforms.

This patch make rimage also get the micro version from SOF CMake, and
write it to the unused hotfix_version field of manifest header.

This update will enable sof_ri_info.py to dump entire file version from
a FW binary for cAVS platforms. So we can check if a FW release candidate
is built from the correct git tag.

Updates from v1:
-  not change definition of any manifest headers to avoid breaking Windows.
-  rimage requires SOF CMake to pass all 3 version fields


RanderWang (approved):
LGTM

aiChaoSONG (approved):
LGTM, but I suggest two things:
- compile sof firmware, and read the firmware with sof_ri_info tool to check if the sof_micro number is really there
- compile sof firmware, and put it to windows to see if there is any error.

mengdonglin:
> LGTM, but I suggest two things:
> 
> * compile sof firmware, and read the firmware with sof_ri_info tool to check if the sof_micro number is really there
Thanks for the reminder :) I've checked this. With this PR and SOF CMake update, sof_ri_info.py can show the micro version properly. Here is the test result:
For v2.1-rc1:
cavs0015 (ADSP Manifest) file offset 0x2300 name ADSPFW build ver 2.1.0.1 feature mask 0xffff image flags 0x0
For v2.1.1:
cavs0015 (ADSP Manifest) file offset 0x2300 name ADSPFW build ver 2.1.1.1 feature mask 0xffff image flags 0x0

> * compile sof firmware, and put it to windows to see if there is any error.
I think this will need your help and confirmation from @RanderWang Thank you!

lgirdwood (approved):

src/rimage.c
```diff
@@ -115,19 +115,21 @@ int main(int argc, char *argv[])
...
 	if (image.fw_ver_string) {
-		ret = sscanf(image.fw_ver_string, "%hu.%hu",
+		ret = sscanf(image.fw_ver_string, "%hu.%hu.%hu",
 			     &image.fw_ver_major,
-			     &image.fw_ver_minor);
+			     &image.fw_ver_minor,
+			     &image.fw_ver_micro);
 
-		if (ret != 2) {
+		if (ret != 3) {
```
marc-hb, line 125: It would be very useful to change the error message a bit. So when there is SOF + rimage mismatch (which _will_ happen in some places) the build failure message is enough to immediately make the difference between an old, 2-fields rimage versus a new, 3 fields rimage. No need to go and try to guess the git versions that were used.

```C
fprintf(stderr,
   "error: cannot parse firmware version major.minor.micro\n");
```

mengdonglin: @marc-hb Thank you! I updated the PR to revise the error message as you suggested. The help message for file version is also updated.

lgirdwood:
@mengdonglin rimage submodule will need an update PR for sof. Thanks

mengdonglin:
> @mengdonglin rimage submodule will need an update PR for sof. Thanks

@lgirdwood Thank you! I submitted PR https://github.com/thesofproject/sof/pull/5791 to update rimage submodule commit for SOF. Please review.

PR #97, opened 2022-06-16
Fix ssl3

juimonen:


aiChaoSONG:
SOF test PR to test this:https://github.com/thesofproject/sof/pull/5928

src/pkcs1_5.c
```diff
@@ -933,6 +896,7 @@ int pkcs_v1_5_verify_man_v2_5(struct image *image,
 		ptr1 - (void *)man, size1, ptr2 - (void *)man, size2);
 #endif
 
+	fprintf(stdout,"paska\n");
```
marc-hb, line 899: I can see you

src/pkcs1_5.c
```diff
@@ -933,6 +886,7 @@ int pkcs_v1_5_verify_man_v2_5(struct image *image,
...
```
plbossart, line 899: spurious edit?

src/pkcs1_5.c
juimonen: heh, yep, fixing my own crap here...

marc-hb:
Thanks @aiChaoSONG for submitting this to CI in test PR https://github.com/thesofproject/sof/pull/5928/.

Everyone note how it's possible to submit an unmerged rimage PR to SOF CI, explanations in the rimage README.

This bug was incredibly time-consuming: after this is merged and well tested over a couple weeks I think it should be backported to SOF release branches (which will require another bit of submodule-fu)

src/pkcs1_5.c
```diff
@@ -624,6 +625,15 @@ int pkcs_v1_5_sign_man_ace_v1_5(struct image *image,
 	EVP_PKEY_free(privkey);
 	return ret;
 }
+#else
+int pkcs_v1_5_sign_man_ace_v1_5(struct image *image,
+			    struct fw_image_manifest_ace_v1_5 *man,
+			    void *ptr1, unsigned int size1, void *ptr2,
+			    unsigned int size2)
+{
+	return -EINVAL;
```
lgirdwood, line 634: @juimonen So IIUC, the code that should be here was removed in the previous patch becuase it was in the wrong place and caused FW to be signed twice ?
@mwasko @marcinszkudlinski it looks like we need the MTL code added here, the code is in the wrong place currently, so not new development just a code move and revalidate. Just checking incase you already have this fix ?

lgirdwood (approved):
We need to take this for v2.2, and we can readd the MTL support in the correct place for v2.3.

src/pkcs1_5.c
juimonen: I had an misunderstanding in my own ver25 ssl3 code (first patch), ace functions we're not implemented for ssl3 at all -> I needed to make quick guards and stub (second patch), because git HEAD was not even compiling in systems with ssl3. So just wanted to fix tgl fast, I need to check the ace code later.

lgirdwood: ok, lets merge this as correct SSL3 code and adapt MTL to follow the same flow when using SSL3.

lgirdwood:
@juimonen can you send a PR to udate teh rimage submodule in SOF. Thanks !

PR #108, opened 2022-09-27
pkcs1_5: set openssl compatibility to v1.1.1

aiChaoSONG:
The openssl and its library shipped with Ubuntu 22.04 is 3.x, which can not be used to sign MTL as the signing function is not migrated to use 3.x API. Set openssl API compatibility to v1.1.1 to workaround signing failure with openssl 3.x on Ubuntu 2204.

Signed-off-by: Chao Song <chao.song@linux.intel.com>


aiChaoSONG:
Jaska will make a PR to add openssl3 support, close this

PR #109, opened 2022-09-27
rimage: make ace15 signing to use cavs25 functions

juimonen:
Ace signing is currently missing openssl3 support, so add it


juimonen:
let's see if this works...

aiChaoSONG:
The build and signing work with this patch on Ubuntu22.04, but I don't have debug key to verify this patch on MTL RVP, let's wait for a while.

aiChaoSONG (approved):
Verified on MTL nocodec RVP, firmware is compiled with my Ubuntu2204

lgirdwood (approved):

lgirdwood:
@juimonen will need a west update now.

PR #113, opened 2022-10-21
rimage: add setting of production bit

juimonen:
When building and signing product images a product bit must be set in manifest headers. Add option -p to rimage to set it.

Co-developed-by:: Adrian Bonislawski <adrian.bonislawski@intel.com>

### DIFF
--- a/src/adsp_config.c
+++ b/src/adsp_config.c
@@ -706,6 +706,7 @@ static void dump_css_v1_5(const struct css_header_v1_5 *css)
 	DUMP_KEY("key_size", "%d", css->key_size);
 	DUMP_KEY("modulus_size", "%d", css->modulus_size);
 	DUMP_KEY("exponent_size", "%d", css->exponent_size);
+	DUMP_KEY("reserved0", "%d", css->reserved0);
 }
 
 static int parse_css_v1_5(const toml_table_t *toml, struct parse_ctx *pctx,
@@ -758,6 +759,10 @@ static int parse_css_v1_5(const toml_table_t *toml, struct parse_ctx *pctx,
 	if (ret < 0)
 		return ret;
 
+	out->reserved0 = parse_uint32_key(css, &ctx, "intel_reserved0", MAN_CSS_INTEL_RES_0, &ret);
+	if (ret < 0)
+		return ret;
+
 	/* check everything parsed */
 	ret = assert_everything_parsed(css, &ctx);
 	if (ret < 0)
@@ -789,6 +794,7 @@ static void dump_css_v1_8(const struct css_header_v1_8 *css)
 	DUMP_KEY("svn", "%d", css->svn);
 	DUMP_KEY("modulus_size", "%d", css->modulus_size);
 	DUMP_KEY("exponent_size", "%d", css->exponent_size);
+	DUMP_KEY("reserved0", "%d", css->reserved0);
 }
 
 static int parse_css_v1_8(const toml_table_t *toml, struct parse_ctx *pctx,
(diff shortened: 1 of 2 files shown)


src/adsp_config.c
```diff
@@ -1023,8 +1037,10 @@ static int parse_signed_pkg(const toml_table_t *toml, struct parse_ctx *pctx,
 	/* bitmap array */
 	bitmap_array = toml_array_in(signed_pkg, "bitmap");
 	if (!bitmap_array) {
-		/* default value */
-		out->bitmap[4] = 8;
+		if (production)
+			out->bitmap[4] = 0x8;
```
abonislawski, line 1041: This is not related to prod bit, please revert all bitmap[] changes

abonislawski (changes requested):
@lgirdwood can you confirm that we can expose reserved0 (production) bit functionality to public?

src/adsp_config.c
abonislawski: or you can add another commit to setup properly IMR type, there is already such option:

fprintf(stdout, "\t -i set IMR type\n");

so you can set bitmap to 0x10 if imr type = 4

src/rimage.c
```diff
@@ -40,11 +40,12 @@ int main(int argc, char *argv[])
 	const char *adsp_config = NULL;
 	int opt, ret, i, first_non_opt;
 	int imr_type = MAN_DEFAULT_IMR_TYPE;
+	bool prod_bit = false;
 	int use_ext_man = 0;
 
 	memset(&image, 0, sizeof(image));
 
-	while ((opt = getopt(argc, argv, "ho:va:s:k:ri:f:b:ec:y:q:")) != -1) {
+	while ((opt = getopt(argc, argv, "ho:va:s:k:ri:f:b:ec:y:q:p")) != -1) {
```
lgirdwood, line 48: Can we not do this in toml i.e. 
```
intel_reserved0 = 0x1234
```

lgirdwood (commented):
Lets make this a toml change

lgirdwood:
> @lgirdwood can you confirm that we can expose reserved0 (production) bit functionality to public?

The naming here is all incorrect, this is for setting IMR bits in the Intel reserved manifest area. Please fix.

src/rimage.c
juimonen: ok

src/adsp_config.c
juimonen: ok

juimonen:
ver2: all stuff coming from toml.

abonislawski:
So there will be another toml config for prod build?

lgirdwood:
@juimonen @abonislawski still needed or problem solved ?

juimonen:
@lgirdwood I guess Adrian was asking "do we need then another toml file for production builds with rimage", and if we do this this way (as you suggested), we will need another toml or you modify toml on the fly when you build. I think Adrians fear was will this make the build "dirty", which I'm not sure. But this is fine for me at least.

lgirdwood:
@juimonen ok, I see - we cant carry this in the toml as the flags are different. Cmd line option it is then.

abonislawski:
@lgirdwood @juimonen can we revert this to cmd line version and merge for v2.4 prod build?

lgirdwood:
@lyakh can you make this into cmd line version and cherry-pick to v2.4. Thanks

lgirdwood:
@juimonen can we close this since we have cmd line version now ? o is any other work needed in main ?

juimonen:
yes closing

PR #116, opened 2022-11-17
rimage: add src support

gongjun-song:
Add src's params to rimage

Signed-off-by: Gongjun Song <gongjun.song@intel.com>


lgirdwood (approved):
@singalsu pls review.

config/tgl-cavs.toml
```diff
@@ -227,3 +227,42 @@ count = 8
...
+		affinity_mask = "0xF"
+		instance_count = "10"
+		domain_types = "1"
+		load_type = "0"
+		module_type = "0x7"
+		auto_start = "0"
+		sched_caps = [1, 0x00008000]
+
+	# pin = [dir, type, sample rate, size, container, channel-cfg]
+	pin = [0, 0, 0xffff, 0xc, 0x8, 0x45ff,
```
singalsu, line 242: Why are these numbers hex? Hex is OK for bit masks and I/O addresses but not nice for numbers like sample rate.

config/tgl-cavs.toml
```diff
@@ -227,3 +227,42 @@ count = 8
...
+		1, 0, 0xf6c9, 0xc, 0x8, 0x45ff]
+
+	# mod_cfg [PAR_0 PAR_1 PAR_2 PAR_3 IS_BYTES CPS IBS OBS MOD_FLAGS CPC OBLS]
+	mod_cfg = [0, 0, 0, 0, 12832, 1365500, 0, 0, 0, 1365, 0,
```
singalsu, line 246: Can the numbers be documented better? I have no idea what are the PAR_X and other acronyms.

RanderWang: No document. It is copied from module spec and used for fw scheduler by windows. It is not supported by cSOF. use it for windows

config/tgl-cavs.toml
gongjun-song: This is used for bit masks. Each bit mask represents whether a certain rate is used. For example, suppose the first bit is to judge whether 16khz is used. Setting 1 means using it and setting 0 means not using it.

singalsu (approved):

PR #117, opened 2022-11-21
Fix module loading issue on windows

RanderWang:
Currently our module type is incorrect. please check https://github.com/thesofproject/rimage/blob/main/src/include/rimage/cavs/cavs_ext_manifest.h#L194-L217

A CCG guy found this bug.


src/adsp_config.c
```diff
@@ -2204,8 +2204,7 @@ static int parse_module(const toml_table_t *toml, struct parse_ctx *pctx,
 		if (strcmp((char *)mod_man->name, "BRNGUP") &&
 		    strcmp((char *)mod_man->name, "BASEFW")) {
 			if (type != i - 1) {
-				log_err(ret, "error: invalid type %d", type);
-				return -EINVAL;
+				log_info(ret, "module type %d is out of order", type);
```
lgirdwood, line 2207: I think it's better to say the "module ID type X is not yet supported. Skipping."

config/mtl.toml
```diff
@@ -192,7 +192,7 @@ count = 15
 	instance_count = "40"
 	domain_types = "0"
 	load_type = "0"
-	module_type = "5"
+	module_type = "4"
```
lgirdwood, line 195: @plbossart fyi - looks like we still need some module type IDs for IPC4 compat but I dont think they are needed by the driver for IPC4 ?

RanderWang: we don't use module type. We set the info in topology, such as uuid and dapm type like dai_in dai_out

src/adsp_config.c
RanderWang: sure

RanderWang:
updated module type for gain, probe and kdtest.

lgirdwood (approved):
@marcinszkudlinski @mwasko pls review.

fkwasowi (approved):

config/mtl.toml
```diff
@@ -288,7 +288,7 @@ count = 15
 	instance_count = "8"
 	domain_types = "0"
 	load_type = "0"
-	module_type = "8"
+	module_type = "12"
```
pblaszko, line 291: These values are interpreted as hex, hence MicSelect module_type should be '0xC'

RanderWang: please check https://github.com/thesofproject/rimage/pull/139

PR #118, opened 2022-12-06
Fix of resource releasing and returning of a error codes

softwarecki:
Fixes errors which came out after static analysis.


marcinszkudlinski (approved):

lgirdwood:
@softwarecki will you update SOF west to include these fixes ?

PR #123, opened 2022-12-23
Add a command line option to set production bit

lyakh:
Add a '-p' command line flag to set Intel reserved bit in IMR area.

Inspired by #113 but unfortunately this is urgent and has to be done with a command line switch, so I had to guess what this should do.

### DIFF
--- a/src/rimage.c
+++ b/src/rimage.c
@@ -40,12 +40,13 @@ int main(int argc, char *argv[])
 	struct adsp *heap_adsp;
 	const char *adsp_config = NULL;
 	int opt, ret, i, first_non_opt;
-	int imr_type = MAN_DEFAULT_IMR_TYPE;
 	int use_ext_man = 0;
 	unsigned int pv_bit = 0;
 	bool imr_type_override = false;
 
 	memset(&image, 0, sizeof(image));
+
+	image.imr_type = MAN_DEFAULT_IMR_TYPE;
 
 	while ((opt = getopt(argc, argv, "ho:va:s:k:ri:f:b:ec:y:q:p")) != -1) {
 		switch (opt) {
@@ -68,7 +69,7 @@ int main(int argc, char *argv[])
 			image.reloc = 1;
 			break;
 		case 'i':
-			imr_type = atoi(optarg);
+			image.imr_type = atoi(optarg);
 			imr_type_override = true;
 			break;
 		case 'f':
@@ -155,7 +156,7 @@ int main(int argc, char *argv[])
 	}
 	image.adsp = heap_adsp;
 	memset(heap_adsp, 0, sizeof(*heap_adsp));
-	ret = adsp_parse_config(adsp_config, heap_adsp, image.verbose);
+	ret = adsp_parse_config(adsp_config, &image);
 	if (ret < 0)
 		goto out;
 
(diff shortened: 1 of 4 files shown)


paulstelian97:
Interesting that you didn't rename the "reserved0" field here. Do we need to do this and do the rename later to have it be done properly? If there's no urgency I'll block it.

lgirdwood (dismissed):
@abonislawski good for you ?

src/rimage.c
```diff
@@ -31,6 +31,7 @@ static void usage(char *name)
 	fprintf(stdout, "\t -e build extended manifest\n");
 	fprintf(stdout, "\t -y verify signed file\n");
 	fprintf(stdout, "\t -q resign binary\n");
+	fprintf(stdout, "\t -p set production bit\n");
```
abonislawski, line 34: please rename to pv bit

src/rimage.c
```diff
@@ -163,15 +168,21 @@ int main(int argc, char *argv[])
 		return resign_image(&image);
 	}
 
-	/* set IMR Type in found machine definition */
-	if (image.adsp->man_v1_8)
+	/* set IMR Type and production bit in found machine definition */
```
abonislawski, line 171: one more field is required to properly setup imr_type, take a look here:

https://github.com/thesofproject/rimage/pull/113#discussion_r1001520828

abonislawski (changes requested):

src/rimage.c
lyakh: you mean the description - "set the PV bit?"

src/rimage.c
lyakh: @abonislawski sorry, don't understand. It does say that the "-i" option already exists. What has to be done?

abonislawski: set bitmap[4] to 0x10 if imr type = 4

Right now we have -i option but imr_type is not set properly, bitmap[4] = always 8 in current code

src/rimage.c
abonislawski: yeap, variables too, everywhere just PV

abonislawski:
Tested on TGL build in sof_ri_info, PV is fine, bitmap is wrong.

Hardcoded will work:
`out->bitmap[4] = 0x10;//adsp->man_v2_5->adsp_file_ext.imr_type == 4 ? 0x10 : 0x8;`

so looks like adsp_file_ext.imr_type is not set yet

lyakh:
> Tested on TGL build in sof_ri_info, PV is fine, bitmap is wrong.
> 
> Hardcoded will work: `out->bitmap[4] = 0x10;//adsp->man_v2_5->adsp_file_ext.imr_type == 4 ? 0x10 : 0x8;`
> 
> so looks like adsp_file_ext.imr_type is not set yet

@abonislawski I presume you did use `-i 4`?

abonislawski:
yes, adsp_file_ext.imr_type is just 0 in this function* so probably it was written too late

*should be at least default 3 (int imr_type = MAN_DEFAULT_IMR_TYPE;)

lyakh:
> yes, adsp_file_ext.imr_type is just 0 in this function* so probably it was written too late
> 
> *should be at least default 3 (int imr_type = MAN_DEFAULT_IMR_TYPE;)

@abonislawski it was indeed. Now it seems to work - tested

abonislawski:
Verified with sof_ri_info and bitmap is correct now but original imr_type is gone:

`image.adsp->man_v2_5->adsp_file_ext.imr_type = imr_type;`

lyakh:
> Verified with sof_ri_info and bitmap is correct now but original imr_type is gone:
> 
> `image.adsp->man_v2_5->adsp_file_ext.imr_type = imr_type;`

@abonislawski this is interesting indeed. I cannot remember any more why I removed these, but looking at it again now it looks like these assignments unconditionally overwrite `imr_type` setting from TOML: https://github.com/thesofproject/rimage/blob/1c48208850c91322b033d5d59fb1a0b4e6648e20/src/adsp_config.c#L1703 - is that correct?

abonislawski:
hmm looks like this code is far from perfect.
Right now we have to set this unconditionally because default value (0) is wrong here:

`out->imr_type = parse_uint32_hex_key(adsp_file_ext, &ctx, "imr_type", 0, &ret);`

so for both cases (default and overwrite) we need to set something outside of this function

There is also a comment for imr_type in parse_adsp_file_ext_v2_5()
```
/*
 * values set in other places in code:
 * - imr_type
 * - comp.limit_offset
 */
```

What you can do:

- bring back unconditional overwrite of imr_type
- or fix default value in parse_uint32_hex_key() (MAN_DEFAULT_IMR_TYPE) and overwrite only if requested

lyakh:
> * fix default value in parse_uint32_hex_key() (MAN_DEFAULT_IMR_TYPE) and overwrite only if requested by rimage parameter

@abonislawski I think this is the better option

abonislawski:
Yes, thats how it should work.

Verified in sof_ri_info, all good.
just please update default value here too:

https://github.com/thesofproject/rimage/blob/1c48208850c91322b033d5d59fb1a0b4e6648e20/src/adsp_config.c#L1583

lyakh:
> Yes, thats how it should work.
> 
> Verified in sof_ri_info, all good. just please update default value here too:
> 
> https://github.com/thesofproject/rimage/blob/1c48208850c91322b033d5d59fb1a0b4e6648e20/src/adsp_config.c#L1583

@abonislawski right, thanks, fixed

abonislawski (approved):

lgirdwood (approved):

PR #124, opened 2023-01-04
config: tgl-cavs: add iir and fir eq module config

juimonen:
This patch adds module config iir eq module.

### DIFF
--- a/config/tgl-cavs.toml
+++ b/config/tgl-cavs.toml
@@ -52,7 +52,7 @@ name = "ADSPFW"
 load_offset = "0x30000"
 
 [module]
-count = 10
+count = 12
 	[[module.entry]]
 	name = "BRNGUP"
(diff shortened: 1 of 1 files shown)


singalsu (approved):

lgirdwood (commented):
LGTM

singalsu (approved):

lgirdwood (approved):

lgirdwood:
@juimonen still draft ?

juimonen:
@lgirdwood we are working with @singalsu to get the eqs up and running. I'll move all related PR's from draft when we get eq working.

juimonen:
@lgirdwood no more draft. I see we have not same modules in tgl-h than in tgl conf... @aiChaoSONG I think smart amp etc should be there also? I can then update eqs also

juimonen:
aah sorry my bad, so I guess the modules are only in tgl-cavs, not in tgl or tgl-h. So this should be good to go.

PR #125, opened 2023-01-11
Ipc4 : add google rtc AEC module support

RanderWang:


config/tgl-cavs.toml
```diff
@@ -286,3 +286,23 @@ count = 10
 
 	# mod_cfg [PAR_0 PAR_1 PAR_2 PAR_3 IS_BYTES CPS IBS OBS MOD_FLAGS CPC OBLS]
 	mod_cfg = [0, 0, 0, 0, 4096, 1000000, 128, 128, 0, 0, 0]
+
+        [[module.entry]]
+        name = "AEC"
```
plbossart, line 291: AEC is a generic name, but there are lots of different versions and implementations. Can we try to have a clearer information on the provider of that solution?

plbossart (changes requested):

config/tgl-cavs.toml
RanderWang: @plbossart  this is define in windows fw manifest and used by windows driver. I use it for compatibility. If we use other name we need to use different UUID

RanderWang:
FW PR: https://github.com/thesofproject/sof/pull/6927

config/tgl-cavs.toml
plbossart: Not talking about the UUID but the 'AEC' name. I am not sure why anyone thought it was a good idea to avoid mentioning this was an Intel IP.

@lgirdwood is this time to enforce a namespace or a convention to avoid issues like this?

plbossart:
> FW PR: [thesofproject/sof#6927](https://github.com/thesofproject/sof/pull/6927)

this is about the Google RTC AEC, what;s the link with the Windows reference you made above @RanderWang ?

RanderWang:
> > FW PR: [thesofproject/sof#6927](https://github.com/thesofproject/sof/pull/6927)
> 
> this is about the Google RTC AEC, what;s the link with the Windows reference you made above @RanderWang ?

Thanks, I got it. I may misunderstand that google AEC is sof AEC. I will add  entry for google AEC

lgirdwood:
@johnylin76 @cujomalainey pls review

config/mtl.toml
```diff
@@ -393,3 +393,23 @@ count = 15
...
+        [[module.entry]]
+        name = "GOOGLE RTC AEC"
+        uuid = "B780A0A6-269F-466F-B477-23DFA05AF758"
+        affinity_mask = "0x3"
+        instance_count = "1"
+        domain_types = "1"
+        load_type = "1"
+        module_type = "10"
+        auto_start = "0"
+        sched_caps = [1, 0x00008000]
```
cujomalainey, line 406: are there any docs anywhere I can look what these fields mean? This is all meaningless to me

config/tgl-cavs.toml
```diff
@@ -286,3 +286,23 @@ count = 10
 
 	# mod_cfg [PAR_0 PAR_1 PAR_2 PAR_3 IS_BYTES CPS IBS OBS MOD_FLAGS CPC OBLS]
 	mod_cfg = [0, 0, 0, 0, 4096, 1000000, 128, 128, 0, 0, 0]
+
+        [[module.entry]]
```
cujomalainey, line 290: do we not have the ability to import a "common" file?

config/tgl-cavs.toml
cujomalainey: @plbossart I think this is in reference to the Google IP AEC

config/mtl.toml
RanderWang: it is converted from aec_module.binmap in module_binmaps for windows

config/tgl-cavs.toml
RanderWang: what's the common part ? Thanks!

cujomalainey: The module itself, i see this file is for tgl only

config/mtl.toml
cujomalainey: I don't think whatever was in there was for our AEC, that would have been intel's or someone elses AEC. We don't deploy on windows

RanderWang: yes, it is not used on windows but it will follow the setting since Linux driver decodes module entry according to windows spec. The used items in module entry are **name and uuid**, others are not used by Linux now.
````
        name = "GOOGLE RTC AEC"  // used by Linux driver
        uuid = "B780A0A6-269F-466F-B477-23DFA05AF758" //use by Linux driver
        affinity_mask = "0x3"  // this module can be used on core 0 and core 1
        instance_count = "1" // only one instance can be enabled
        domain_types = "1"  // DP schedule domain
        load_type = "1" //  loadable module
        module_type = "10" // AEC module
        auto_start = "0" 
        sched_caps = [1, 0x00008000] 

        # pin = [dir, type, sample rate, size, container, channel-cfg]
        pin = [0, 0, 0x8, 0x2, 0x2, 0x1,    // support input format of 16 bit, 1 ch, 48000k
                0, 0, 0x8, 0x2, 0x2, 0x4,      // support input format of 16bit  2 ch, 48000k
                1, 0, 0x8, 0x2, 0x2, 0x1]    // support output format of 16bit  1 ch, 48000k

        # mod_cfg [PAR_0 PAR_1 PAR_2 PAR_3 IS_BYTES CPS IBS OBS MOD_FLAGS CPC OBLS]
        mod_cfg = [0, 0, 0, 0, 49152, 417000, 16, 16, 0, 2085, 0,     // unused by Linux driver
                   1, 0, 0, 0, 49152, 29525000, 16, 16, 0, 147625, 0]
````

Comment (RanderWang) [reply config/mtl.toml:406]: Ipc4 fw will be packed as a binary can work on both windows and Linux platforms  and the packed format was defined first on windows, so we have to define any module in windows layout.

Comment (cujomalainey) [reply config/mtl.toml:406]: Assuming this is for our module, I can't sign off until I get detailed docs explaining the range values of these fields, are they going to be added anywhere?

Comment (RanderWang) [reply config/mtl.toml:406]: sure, let's add it in my shop list

Comment (kv2019i) [reply config/mtl.toml:406]: Ack @cujomalainey  We have documentation in the rimage headers (e.g. in https://github.com/thesofproject/rimage/blob/main/src/include/rimage/sof/user/manifest.h ) and we should add any missing pieces here so that this is standalone and sufficient to fill in any needed data to describe the modules. FYI to @mwasko as well, we need to cover this in sof-docs for doing new modules.

Comment (RanderWang) [reply config/mtl.toml:406]: @kv2019i  please check https://github.com/thesofproject/sof-docs/pull/451. Thanks!

**Comment by RanderWang** [comment]
rebase to latest code

**Comment by lgirdwood** [comment]
@plbossart update good for you ?

**Review by lgirdwood — DISMISSED** [review]

**Comment by lgirdwood** [comment]
@RanderWang conflict.

**Comment by RanderWang** [comment]
> @RanderWang conflict.

Thanks, updated

**Review by lgirdwood — APPROVED** [review]

**Review by cujomalainey — COMMENTED** [review]
whoops forgot to publish these on the previous version

File: config/mtl.toml [hunk 4]
```diff
@@ -413,3 +413,19 @@ count = 16
 
 	# mod_cfg [PAR_0 PAR_1 PAR_2 PAR_3 IS_BYTES CPS IBS OBS MOD_FLAGS CPC OBLS]
 	mod_cfg = [0, 0, 0, 0, 4096, 1000000, 128, 128, 0, 0, 0]
+
+        [[module.entry]]
+        name = "RTC_AEC"
+        uuid = "B780A0A6-269F-466F-B477-23DFA05AF758"
+        affinity_mask = "0x3"
+        instance_count = "1"
```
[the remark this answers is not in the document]
RanderWang: ok, let's use 2

config/mtl.toml
```diff
@@ -413,3 +413,19 @@ count = 16
 
 	# mod_cfg [PAR_0 PAR_1 PAR_2 PAR_3 IS_BYTES CPS IBS OBS MOD_FLAGS CPC OBLS]
 	mod_cfg = [0, 0, 0, 0, 4096, 1000000, 128, 128, 0, 0, 0]
+
+        [[module.entry]]
+        name = "RTC_AEC"
+        uuid = "B780A0A6-269F-466F-B477-23DFA05AF758"
+        affinity_mask = "0x3"
```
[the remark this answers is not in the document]
RanderWang: It is a duplication of general AEC. Let's use 0xF for RTC AEC

config/mtl.toml
```diff
@@ -413,3 +413,19 @@ count = 16
...
+        instance_count = "1"
+        domain_types = "1"
```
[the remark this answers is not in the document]
RanderWang: complex data processing module.  ll means low latency module like copier, volume, mixer for simple data processing.

config/mtl.toml
```diff
@@ -413,3 +413,19 @@ count = 16
...
+        load_type = "1"
```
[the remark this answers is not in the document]
RanderWang: sure

RanderWang:
@cujomalainey  sorry, update late.

config/mtl.toml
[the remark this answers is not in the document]
mwasko: @cujomalainey details on DP tasks description can be found here: https://thesofproject.github.io/latest/architectures/firmware/sof-zephyr/mpp_layer/mpp_scheduling.html

config/mtl.toml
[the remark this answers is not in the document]
lgirdwood: @mwasko could you clarify here, if the ext manifest says only 1 module instance but topology says 2 what should happen ?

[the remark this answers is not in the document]
mwasko: > @mwasko could you clarify here, if the ext manifest says only 1 module instance but topology says 2 what should happen ?

In such case FW should report an error on attempt to create second module instance.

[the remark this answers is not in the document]
lgirdwood: ok, wont this mean generating differnt baseFWs for different topologies ?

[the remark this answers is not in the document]
cujomalainey: What is the concern of setting this to unlimited?

config/tgl-cavs.toml
```diff
@@ -388,3 +388,18 @@ count = 15
 	mod_cfg = [0, 0, 0, 0, 960, 488500, 16, 16, 0, 0, 0,
 			1, 0, 0, 0, 960, 964500, 16, 16, 0, 0, 0,
 			2, 0, 0, 0, 960, 2003000, 16, 16, 0, 0, 0]
+        [[module.entry]]
+        name = "RTC_AEC"
+        uuid = "B780A0A6-269F-466F-B477-23DFA05AF758"
+        affinity_mask = "0xF"
+        instance_count = "2"
```
cujomalainey, line 395: I still fail to understand why this is a mandatory field. In a truly versatile firmware this should only be enforced when needed? And this also implies the topology cannot be trusted.

cujomalainey (changes requested):
Also this parser is impossible to read at face value without the docs with you. Would be great to make the parser smarter to handle enums

config/tgl-cavs.toml
mwasko: @cujomalainey please check my comment above. Generally for 3rd party components we can make it optional with assumption that if no instance_count is defined then it is unlimited.

config/mtl.toml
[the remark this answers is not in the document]
plbossart: @mwasko @lgirdwood can I ask a somewhat unrelated question related to the affinity_mask? Is there a good reason why some modules are restricted to core0, some to core0/1, and most do not have any restrictions.

I am specifically concerned about the core0 restriction, which I view as incompatible with the direction to allow low-latency pipelines to run on all cores. if a mixin is always on core0, then something's not adding up.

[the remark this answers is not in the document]
cujomalainey: @plbossart agreed, this seems to put component level restrictions on topology configuration without context. I would expect more restrictions like "MIPS estimates cannot violate this amount per core in worst case scenarios" and the stack check for this platform.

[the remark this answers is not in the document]
mwasko: > @mwasko @lgirdwood can I ask a somewhat unrelated question related to the affinity_mask? Is there a good reason why some modules are restricted to core0, some to core0/1, and most do not have any restrictions.

This field is used by Windows driver that has more flexible approach on how pipelines and modules are instantiated. The pipeline/component is assigned to specific core based on current CPU resource availability (MCPS). 

> I am specifically concerned about the core0 restriction, which I view as incompatible with the direction to allow low-latency pipelines to run on all cores. if a mixin is always on core0, then something's not adding up.

Originally Windows was grouping LL modules on Core 0 and I believe that is why Rander defined it like that. However, @plbossart is right here and with SOF we allow low-latency pipelines to run on all cores so all the built-in modules should not have affinity restrictions. The 3rd party modules affinity is up to integrator but the default option should be 'no restrictions'.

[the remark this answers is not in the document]
plbossart: Thanks @mwasko, your explanations match the little understanding I have of the modules.

@RanderWang can we have a follow-up patch to clean-up the affinity definitions for built-in modules, i.e. remove the restriction to core0 and core0-1 for all built-in modules? Thanks!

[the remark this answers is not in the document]
RanderWang: sure, I will follow up it.

PR #126, opened 2023-01-27
Convert all ELF addresses to cached for calculations

lyakh:
Rimage calculates sizes of ELF sections, for which it has to use addresses from the same address space: either all cached or all uncached. The ELF image itself can contain mixed addresses. Convert all to cached for internal calculations.

### DIFF
--- a/src/elf.c
+++ b/src/elf.c
@@ -12,6 +12,11 @@
 #include <rimage/cse.h>
 #include <rimage/manifest.h>
 
+static unsigned long uncache_to_cache(const struct image *image, unsigned long address)
+{
+	return ((address & ~image->adsp->alias_mask) | image->adsp->alias_cached);
+}
+
 static int elf_read_sections(struct image *image, struct module *module,
 			     int module_index)
 {
@@ -110,6 +115,7 @@ static int elf_read_sections(struct image *image, struct module *module,
 			continue;
 		}
 
+		section[i].vaddr = uncache_to_cache(image, section[i].vaddr);
 		module->num_sections++;
 
 		if (!image->verbose)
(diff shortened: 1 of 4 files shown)


paulstelian97:
Merge commit, which I'd consider a mess. That said I don't know if the rule that rebase must be preferred to merge also applies to this repo.

lyakh:
If this method is acceptable, I'll add other platforms too

lyakh:
addresses https://github.com/thesofproject/sof/issues/6741

lyakh:
I marked it "ready for review" to actually have it reviewed, but before merging or at least before updating SOF to use the updated rimage, we have to add all other platforms too

src/elf.c
```diff
@@ -12,6 +12,11 @@
 #include <rimage/cse.h>
 #include <rimage/manifest.h>
 
+static unsigned long uncache_to_cache(const struct image *image, unsigned long address)
+{
+	return ((address & ~(7 << 29)) | image->adsp->alias_cached);
```
lgirdwood, line 17: This is quite Intel specific, could be an op based on platform.

lgirdwood:
@lyakh is there a Zephyr PR for removal of the python script or do we need other SOF PRs merged first ?

lyakh:
> @lyakh is there a Zephyr PR for removal of the python script or do we need other SOF PRs merged first ?

@lgirdwood we can merge this first. As we see, all tests pass. Running that .py script on top of this change shouldn't hurt. After this is merged then we can remove that python script

lgirdwood:
@lyakh just a few review comments then we are good.

lgirdwood (approved):

kv2019i (approved):
Looks good to me.

Is the plan to add other platforms later? I think this can be merged now, but to get rid of the python script, we need to have all platforms updated with the cached/uncached split info.

PR #129, opened 2023-02-01
loadable modules: add buildin api info structure

pjdobrowolski:
Changes in manifest structures with additional info about buildin api. That informations are available in iadk modules and are already implemented.

### DIFF
--- a/src/include/rimage/sof/user/manifest.h
+++ b/src/include/rimage/sof/user/manifest.h
@@ -226,6 +226,27 @@ struct sof_man_module_manifest {
 };
 
 /*
+ *  Api version for sof loadable modules
+ */
+typedef union sof_module_api_version
+{
+    uint32_t full;
+    struct
+    {
+        uint32_t minor    : 10;
+        uint32_t middle   : 10;
+        uint32_t major    : 10;
+        uint32_t reserved : 2;
+    } fields;
+} sof_module_api_version;
+
+typedef const struct
+{
+    uint32_t FORMAT;
+    sof_module_api_version API_VERSION_NUMBER;
+} sof_module_build_info;
+
+/*
  * Module offset in manifest.
  */
 #define SOF_MAN_MODULE_OFFSET(index) \


src/include/rimage/sof/user/manifest.h
```diff
@@ -26,13 +26,20 @@
...
 	uint32_t load_type:4;	/* SOF_MAN_MOD_TYPE_ */
 	uint32_t auto_start:1;
 	uint32_t domain_ll:1;
 	uint32_t domain_dp:1;
 	uint32_t lib_code:1;
-	uint32_t rsvd_:24;
+	uint32_t domain_rtos:1;
+	uint32_t core_type:8;
+	uint32_t lib_type:1;
+	uint32_t rsvd_:14;
```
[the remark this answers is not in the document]
pjdobrowolski: Agree. I will split each bit field into separate commit with description. It might clear up things for now.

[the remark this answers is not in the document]
pjdobrowolski: Changed it to much simpler way of detecting API of loadable modules basing on buildin api info.

src/include/rimage/sof/user/manifest.h
```diff
@@ -225,6 +225,27 @@ struct sof_man_module_manifest {
...
+        uint32_t reserved : 2;
+    } fields;
+} sof_module_api_version;
+
+typedef const struct
+{
+    uint32_t FORMAT;
+    sof_module_api_version API_VERSION_NUMBER;
+} sof_module_build_info;
+
```
lgirdwood, line 248: Any reason for typedefs and member names in capitals ? Btw, we will need to pack these.

pjdobrowolski: This structure is used in iadk binaries also. Will packing impact closed source?

softwarecki:
I think that this structure should be placed in sof or zephyr. It has no relation to rimage, and will not be used by it.

src/include/rimage/sof/user/manifest.h
lgirdwood: ok, if it needs to go in main FW tree pls close here. It will need to be packed so all compilers treat it the same (i.e. there is no ambiguous sizing).

PR #130, opened 2023-02-08
Add Init config type for modules

ranj063:
USe some reverved bits to add the init config information for modules that will be used to indicate whether the base config extension needs to be added to the init module payload or not.

### DIFF
--- a/src/include/rimage/sof/user/manifest.h
+++ b/src/include/rimage/sof/user/manifest.h
@@ -26,13 +26,18 @@
 #define SOF_MAN_MOD_TYPE_BUILTIN	0
 #define SOF_MAN_MOD_TYPE_MODULE		1
 
+/* module init config */
+#define SOF_MAN_MOD_INIT_CONFIG_BASE_CFG               0 /* Base config only */
+#define SOF_MAN_MOD_INIT_CONFIG_BASE_CFG_WITH_EXT      1 /* Base config with extension */
+
 struct sof_man_module_type {
 	uint32_t load_type:4;	/* SOF_MAN_MOD_TYPE_ */
 	uint32_t auto_start:1;
 	uint32_t domain_ll:1;
 	uint32_t domain_dp:1;
 	uint32_t lib_code:1;
-	uint32_t rsvd_:24;
+	uint32_t init_config:4; /* SOF_MAN_MOD_INIT_CONFIG_ */
+	uint32_t rsvd_:20;
 };
 
 /* segment flags.type */
--- a/config/mtl.toml
+++ b/config/mtl.toml
@@ -402,6 +402,7 @@ count = 16
 	instance_count = "1"
 	domain_types = "0"
 	load_type = "0"
+	init_config = "1"
 	module_type = "13"
 	auto_start = "0"
 	sched_caps = [1, 0x00008000]
--- a/config/tgl-cavs.toml
+++ b/config/tgl-cavs.toml
@@ -276,6 +276,7 @@ count = 12
 	domain_types = "0"
 	load_type = "0"
 	module_type = "13"
+	init_config = "1"
 	auto_start = "0"
 	sched_caps = [1, 0x00008000]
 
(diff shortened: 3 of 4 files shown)


aiChaoSONG (approved):

ujfalusi (approved):

src/include/rimage/sof/user/manifest.h
```diff
@@ -26,13 +26,18 @@
...
+
 struct sof_man_module_type {
 	uint32_t load_type:4;	/* SOF_MAN_MOD_TYPE_ */
 	uint32_t auto_start:1;
 	uint32_t domain_ll:1;
 	uint32_t domain_dp:1;
 	uint32_t lib_code:1;
-	uint32_t rsvd_:24;
+	uint32_t init_config:4; /* SOF_MAN_MOD_INIT_CONFIG_ */
+	uint32_t rsvd_:20;
```
kv2019i, line 40: This needs to be added to sof-docs as well (follow-up to https://github.com/thesofproject/sof-docs/pull/451 ).

kv2019i (approved):
Thanks, looks good!

pblaszko:
This implementation does not break anything itself, as it just reuses some reserved bits of module type in manifest.
But what is a purpose of this change? Modules with only base_cfg or with base_cfg+base_cfg_extension are only a subset of overall FW modules. Many infrastructure modules have more data in payload. Is your change intended for modules built with IADK?

kv2019i:
@pblaszko wrote:
> This implementation does not break anything itself, as it just reuses some reserved bits of module type in manifest. But what is > a purpose of this change? Modules with only base_cfg or with base_cfg+base_cfg_extension are only a subset of overall FW 
> modules. Many infrastructure modules have more data in payload. Is your change intended for modules built with IADK?

This is related to kernel PR 4160 and specifically this comment https://github.com/thesofproject/linux/pull/4160#discussion_r1090912209 .

AFAIU, for base modules (like copier) the kernel can have hard-coded knowledge of each module, but for generic processing modules (like IADK ones), it's not scalable to need separate changes to Linux for each new type of module. I.e. the manifest needs to tell kernel if a module absolutely needs the base-ext to function correctly (smart amp is one example). Our earlier alternative was to use the topology file, but during the discussion it became apparent this is not the correct place to store the meta information.

This of course leaves still the case where a module has a completely custom initialization interface and it needs the complete set to function correctly. For these, the Linux kernel has to be modified to support the module. This is possible, but a very slow to deploy path -- this should really only be used for base modules such as copier.

pblaszko:
> @pblaszko wrote:
> 
> > This implementation does not break anything itself, as it just reuses some reserved bits of module type in manifest. But what is > a purpose of this change? Modules with only base_cfg or with base_cfg+base_cfg_extension are only a subset of overall FW
> > modules. Many infrastructure modules have more data in payload. Is your change intended for modules built with IADK?
> 
> This is related to kernel PR 4160 and specifically this comment [thesofproject/linux#4160 (comment)](https://github.com/thesofproject/linux/pull/4160#discussion_r1090912209) .
> 
> AFAIU, for base modules (like copier) the kernel can have hard-coded knowledge of each module, but for generic processing modules (like IADK ones), it's not scalable to need separate changes to Linux for each new type of module. I.e. the manifest needs to tell kernel if a module absolutely needs the base-ext to function correctly (smart amp is one example). Our earlier alternative was to use the topology file, but during the discussion it became apparent this is not the correct place to store the meta information.
> 
> This of course leaves still the case where a module has a completely custom initialization interface and it needs the complete set to function correctly. For these, the Linux kernel has to be modified to support the module. This is possible, but a very slow to deploy path -- this should really only be used for base modules such as copier.

Thank you @kv2019i !
So this will be an optimization for modules that fit into specified structures, but other modules with different structure can still be used (preferably only FW infrastructure). This is fine for me.

pblaszko (approved):

src/include/rimage/sof/user/manifest.h
lgirdwood: ack - lets get it in the docs.

lgirdwood (approved):

PR #131, opened 2023-02-09
Fix regression - make default return code an error

lyakh:
A recent commit ec649f37d661 ("Convert all ELF addresses to cached for calculations") had a bug: it assumed a call to parse_uint32_hex_key() for an absent key would return an error by default, which isn't the case. To force it set the default return code to an error value.

### DIFF
--- a/config/mtl.toml
+++ b/config/mtl.toml
@@ -3,6 +3,7 @@ version = [3, 0]
 [adsp]
 name = "mtl"
 image_size = "0x2C0000" # (22) bank * 128KB
+alias_mask = "0xE0000000"
 
 [[adsp.mem_zone]]
 type = "ROM"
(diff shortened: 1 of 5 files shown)


lgirdwood (approved):

lgirdwood:
@kv2019i fyi.

lyakh:
This one passed testing too: https://github.com/thesofproject/sof/pull/7062 (before the addition of the 2nd commit though)

lyakh:
https://github.com/thesofproject/sof/pull/7062 relaunched and passed all tests with both commits too

ranj063:
@lyakh this still fails with xtos builds.

marc-hb:
@lyakh nice try but I think your test PR https://github.com/thesofproject/sof/pull/7062 did not work because it updated the west manifest only, not the submodules. See comment in sof/west.yml.

We must decouple submodules and west but  this will be somewhat disruptive for CIs and there's always some higher priority than breaking CI. Maybe after 2.5

lyakh:
> @lyakh nice try but I think your test PR [thesofproject/sof#7062](https://github.com/thesofproject/sof/pull/7062) did not work because it updated the west manifest only, not the submodules. See comment in sof/west.yml.
> 
> We must decouple submodules and west but this will be somewhat disruptive for CIs and there's always some higher priority than breaking CI. Maybe after 2.5

@marc-hb yes, thanks, we figured it out last week, trying to fix it for all configurations...

PR #132, opened 2023-02-10
config:tgl/tglh: Do not set cached/uncached address aliases

ranj063:
tgl.toml and tgl-h.toml are only used for xtos builds.

Suggested-by: Guennadi Liakhovetski <guennadi.liakhovetski@linux.intel.com>
Signed-off-by: Ranjani Sridharan <ranjani.sridharan@linux.intel.com>

### DIFF
--- a/config/tgl-h.toml
+++ b/config/tgl-h.toml
@@ -3,7 +3,6 @@ version = [2, 5]
 [adsp]
 name = "tgl"
 image_size = "0x1F0000" # (30 + 1) bank * 64KB
-alias_mask = "0xE0000000"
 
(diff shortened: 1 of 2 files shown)


lyakh (approved):
assuming these two are never used with Zephyr

kv2019i:
@lyakh IPC3 + TGL is no longer a supported config, so we can let this pass, but potentially more fixups are needed later.

kv2019i (approved):

kv2019i:
If this doesn't work out, then we have to revert the whole change to unblock more urgent rimage changes.

PR #133, opened 2023-02-13
Don't convert ROM addresses to cached aliases

lyakh:
We don't send ROM data to the DSP so there's no need to convert ROM addresses to cached aliases. Doing so breaks calculations with IPC4 XTOS builds.

### DIFF
--- a/src/elf.c
+++ b/src/elf.c
@@ -14,7 +14,7 @@
 
 static unsigned long uncache_to_cache(const struct image *image, unsigned long address)
 {
-	return ((address & ~image->adsp->alias_mask) | image->adsp->alias_cached);
+	return (address & ~image->adsp->alias_mask) | image->adsp->alias_cached;
 }
 
 static int elf_read_sections(struct image *image, struct module *module,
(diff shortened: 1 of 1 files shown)


src/elf.c
```diff
@@ -14,7 +14,14 @@
 
 static unsigned long uncache_to_cache(const struct image *image, unsigned long address)
 {
-	return ((address & ~image->adsp->alias_mask) | image->adsp->alias_cached);
+	unsigned long base = image->adsp->mem_zones[SOF_FW_BLK_TYPE_ROM].base;
+	size_t size = image->adsp->mem_zones[SOF_FW_BLK_TYPE_ROM].size;
```
kv2019i, line 18: Hmm, I think a bit of background would be needed either in commit or code comments. To start with, I don't fully understand "we don't send ROM data to DSP"  in this context. I mean surely the DSP is consuming ROM data as well, it's just we don't put the put ROM data into the FW main image (that we are producing here). So maybe just rephrase a bit.

Then I'm wondering could this ROM check be at a higher level. This is a bit surprising that the uncache_to_cache() has this logic inside. But I guess you looked at different options and identifying the ROM bits is not doable in the caller...?


But then I wonder could this check be on a higher level

kv2019i (commented):
A few comments. I guess this is ok still.

kv2019i (approved):
Thanks, much easier to follow!

PR #137, opened 2023-02-22
file_utils: manifest: elf: Separation of the code fragments to the new functions

softwarecki:
A new function create_file_name has been created that prepares the name of the output file with the given extensions. Unlike the previous solution, it checks whether the new name will fit in the output buffer.

A new file_utils.c file has been prepared, into which generic functions to simplify the use of files will be moved.

Added a new get_file_size function which allows to conveniently get the size of the file.
Simplified retrieval of file size by using the function get_file_size in manifest and elf.

### DIFF
--- a/src/ext_manifest.c
+++ b/src/ext_manifest.c
@@ -14,6 +14,7 @@
 #include <rimage/rimage.h>
 #include <rimage/cavs/cavs_ext_manifest.h>
 #include <rimage/manifest.h>
+#include <rimage/file_utils.h>
 
 const struct ext_man_header ext_man_template = {
 	.magic = EXT_MAN_MAGIC_NUMBER,
@@ -24,10 +25,14 @@ const struct ext_man_header ext_man_template = {
 
 static int ext_man_open_file(struct image *image)
 {
-	/* open extended manifest outfile for writing */
-	sprintf(image->out_ext_man_file, "%s.xman", image->out_file);
-	unlink(image->out_ext_man_file);
+	int ret;
 
+	ret = create_file_name(image->out_ext_man_file, sizeof(image->out_ext_man_file),
+			       image->out_file, "xman");
+	if (ret)
+		return ret;
+
+	/* open extended manifest outfile for writing */
 	image->out_ext_man_fd = fopen(image->out_ext_man_file, "wb");
 	if (!image->out_ext_man_fd) {
 		fprintf(stderr, "error: unable to open %s for writing %d\n",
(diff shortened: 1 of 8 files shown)


src/include/rimage/file_utils.h
```diff
@@ -0,0 +1,22 @@
...
+ * @param new_name char[] destination of new file name
+ * @param name_size new file name buffer capacity
+ * @param template_name File name used as a template for the new name
+ * @param new_ext extension of the new file name
+ * @param error code, 0 when success
+ */
+int create_file_name(char *new_name, const size_t name_size, const char *template_name,
+		     const char *new_ext);
+
+#endif /* __FILE_UTILS_H__ */
```
lgirdwood, line 31: missing a new line

src/ext_manifest.c
```diff
@@ -24,10 +25,14 @@ const struct ext_man_header ext_man_template = {
 
 static int ext_man_open_file(struct image *image)
 {
-	/* open extended manifest outfile for writing */
-	sprintf(image->out_ext_man_file, "%s.xman", image->out_file);
-	unlink(image->out_ext_man_file);
```
lgirdwood, line 29: Dont we need to delete the old file ?

softwarecki: Old file is deleted by `create_file_name` function.

pjdobrowolski (approved):

src/file_utils.c
```diff
@@ -33,3 +34,38 @@ int create_file_name(char *new_name, const size_t name_size, const char *templat
 
 	return 0;
 }
+
+/**
```
marcinszkudlinski, line 38: double comments - in C and in H
also, maybe add 
@param **[in]**  f
@param **[out]** size output for file size

and sort - input params at start, ouput at the end. I'm not sure if we do have it in coding standard. but keeping output params at end of the params list makes the code more readable

src/file_utils.c
```diff
@@ -36,3 +36,38 @@ int create_file_name(char *new_name, const size_t name_size, const char *templat
...
+int get_file_size(FILE *f, const char* filename, size_t *size)
+{
+	int ret;
+
+	assert(size);
+
+	/* get file size */
+	ret = fseek(f, 0, SEEK_END);
+	if (ret < 0) {
+		fprintf(stderr, "error: unable to seek eof %s %d\n", filename, errno);
```
lgirdwood, line 56: fwiw, its really nice to use ```strerror(errno)``` in places like this. i.e. instead of printing the %d errno you can print the %s strerror(errno)

lgirdwood (approved):

aborisovich (approved):

PR #138, opened 2023-02-23
toml_utils: adsp_config: Moved generic parser functions to toml_utils

softwarecki:
A set of functions which make it easier to parse toml files has been separated into a new file. In the near future I will add a bunch of new parsing functions. It is good to have prepared a place for them.


lgirdwood (approved):
Fwiw, git sometimes needs a cmd line hint when moving code as it sometimes cant detect a code move.

pblaszko (approved):

PR #139, opened 2023-02-24
rimage: use hex number

RanderWang:
Module type is interpreted as hex number

### DIFF
--- a/config/mtl.toml
+++ b/config/mtl.toml
@@ -297,7 +297,7 @@ count = 16
 	domain_types = "0"
 	load_type = "0"
 	init_config = "1"
-	module_type = "12"
+	module_type = "0xC"
 	auto_start = "0"
 	sched_caps = [1, 0x00008000]
 
@@ -412,7 +412,7 @@ count = 16
 	domain_types = "0"
 	load_type = "0"
 	init_config = "1"
-	module_type = "13"
+	module_type = "0xD"
 	auto_start = "0"
 	sched_caps = [1, 0x00008000]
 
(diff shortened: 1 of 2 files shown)


RanderWang:
fix: https://github.com/thesofproject/rimage/pull/117#discussion_r1115960900 "These values are interpreted as hex, hence MicSelect module_type should be '0xC'"

pblaszko (approved):
Thank you for so fast response and fix!

KokoSoft:
Its not a fix. It is a dirty work around. The toml parser used in this project supports hexadecimal numbers. Someone, for unknown reasons, wrote them as strings (in " ") and made their own hex number parsing function. Instead, we need to fix the `parse_uint32_hex_key` function to check if the value starts with 0x. The best is to fix all numeric values in a toml files, and write them as numbers, not strings. This also will require fix the mentioned function, to accept integers values.

pblaszko:
> Its not a fix. It is a dirty work around. The toml parser used in this project supports hexadecimal numbers. Someone, for unknown reasons, wrote them as strings (in " ") and made their own hex number parsing function. Instead, we need to fix the `parse_uint32_hex_key` function to check if the value starts with 0x. The best is to fix all numeric values in a toml files, and write them as numbers, not strings. This also will require fix the mentioned function, to accept integers values.

Thank you for mentioning it is not a proper fix.
On the other hand, other modules in the same file already use the same work-around, so this patch does not make it any worse.
As we need this change urgently, I propose to merge this change and open an issue for rimage, to plan proper fix in the future.

iganakov (approved):

softwarecki:
Instead of correcting every occurrence in the toml file, he suggests solving the problem at the source. This will avoid similar problems in the future.  https://github.com/thesofproject/rimage/pull/140

PR #141, opened 2023-02-27
Set of a new hash computing functions (SP1)

softwarecki:
Moved byte_swap function to a new misc_utils.c file. This function will be used in different places in the code.

Prepared a set of fa new unctions for computing digest with error handling. The hash context has been placed into a separate structure. This allows to conveniently pass the calculated digest along with its length and used algorithm.

Switched to use new hash functions. The parameter list for the rimage_sign and rimage_verify functions has been simplified. The necessary information is now included in the hash_context structure. In addition, code fragments that may be common to different versions of the manifest have been separated.

Fields related to hash computing have been removed from the image structure. Removed openssl headers from rimage.h, which made it necessary to add some missing includes in several files.


aborisovich (changes requested):
I'll just skip repeating what cppcheck said and idicate, that the fact that it is "green" does not mean it did not find issues ;-)

src/manifest.c
```diff
@@ -1455,35 +1486,43 @@ int man_write_fw_ace_v1_5(struct image *image)
...
+	bytes_swap(m->signed_pkg.module[0].hash, sizeof(m->signed_pkg.module[0].hash));
 
 	/* calculate hash - SHA384 on CAVS2_5+ */
-	module_sha384_create(image);
-	module_sha_update(image, image->fw_image,
-			  sizeof(struct CsePartitionDirHeader_v2_5) +
-			  sizeof(struct CsePartitionDirEntry) * 3);
-	module_sha_update(image, image->fw_image + 0x4c0, image->image_end - 0x4c0);
-	module_sha_complete(image, hash);
+	hash_sha384_init(&hash);
```
pblaszko, line 1513: check error code

src/manifest.c
```diff
@@ -1455,35 +1486,43 @@ int man_write_fw_ace_v1_5(struct image *image)
...
+	hash_update(&hash, image->fw_image,
```
pblaszko, line 1514: check error code

src/manifest.c
```diff
@@ -1455,35 +1486,43 @@ int man_write_fw_ace_v1_5(struct image *image)
...
+		    sizeof(struct CsePartitionDirHeader_v2_5) +
+		    sizeof(struct CsePartitionDirEntry) * 3);
+
+	hash_update(&hash, image->fw_image + 0x4c0, image->image_end - 0x4c0);
```
pblaszko, line 1518: check error code

src/manifest.c
```diff
@@ -1455,35 +1486,43 @@ int man_write_fw_ace_v1_5(struct image *image)
...
+	hash_finalize(&hash);
```
pblaszko, line 1519: check ec

src/pkcs1_5.c
```diff
@@ -418,22 +400,20 @@ int pkcs_v1_5_sign_man_v1_5(struct image *image,
 	}
 
 	/* calculate the digest */
-	module_sha256_create(image);
-	module_sha_update(image, ptr1, size1);
-	module_sha_complete(image, digest);
+	hash_sha256_init(&digest);
```
pblaszko, line 403: check ec

src/pkcs1_5.c
```diff
@@ -418,22 +400,20 @@ int pkcs_v1_5_sign_man_v1_5(struct image *image,
...
+	hash_update(&digest, ptr1, size1);
```
pblaszko, line 404: check ec

src/pkcs1_5.c
```diff
@@ -487,22 +468,21 @@ int pkcs_v1_5_sign_man_v1_8(struct image *image,
 	}
 
 	/* calculate the digest */
-	module_sha256_create(image);
-	module_sha_update(image, ptr1, size1);
-	module_sha_update(image, ptr2, size2);
-	module_sha_complete(image, digest);
+	hash_sha256_init(&digest);
```
pblaszko, line 471: check ec

src/pkcs1_5.c
```diff
@@ -487,22 +468,21 @@ int pkcs_v1_5_sign_man_v1_8(struct image *image,
...
+	hash_update(&digest, ptr1, size1);
```
pblaszko, line 472: check ec

src/pkcs1_5.c
```diff
@@ -487,22 +468,21 @@ int pkcs_v1_5_sign_man_v1_8(struct image *image,
...
+	hash_update(&digest, ptr2, size2);
```
pblaszko, line 473: check ec

src/pkcs1_5.c
```diff
@@ -547,22 +528,21 @@ int pkcs_v1_5_sign_man_v2_5(struct image *image,
 	}
 
 	/* calculate the digest - SHA384 on CAVS2_5+ */
-	module_sha384_create(image);
-	module_sha_update(image, ptr1, size1);
-	module_sha_update(image, ptr2, size2);
-	module_sha_complete(image, digest);
+	hash_sha384_init(&digest);
```
pblaszko, line 531: check ec

src/pkcs1_5.c
```diff
@@ -547,22 +528,21 @@ int pkcs_v1_5_sign_man_v2_5(struct image *image,
...
+	hash_update(&digest, ptr1, size1);
```
pblaszko, line 532: check ec

src/pkcs1_5.c
```diff
@@ -547,22 +528,21 @@ int pkcs_v1_5_sign_man_v2_5(struct image *image,
...
+	hash_update(&digest, ptr2, size2);
```
pblaszko, line 533: check ec

src/pkcs1_5.c
```diff
@@ -607,22 +588,21 @@ int pkcs_v1_5_sign_man_ace_v1_5(struct image *image,
...
```
pblaszko, line 593: check ec... and for all next occurencies :)

softwarecki: The error code is stored in hash_context. You only need to check it once at the end

pblaszko: So if you call 3 functions in a row and then check error code, do you know which function failed? Error reporting becomes less readable and precise.
Is it always allowed to call next function if previous one failed?

src/misc_utils.c
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright(c) 2018-2023 Intel Corporation. All rights reserved.
+ *
+ * Author: Liam Girdwood <liam.r.girdwood@linux.intel.com>
```
marcinszkudlinski, line 5: are you Liam?

src/hash.c
```diff
@@ -46,62 +49,157 @@ void EVP_MD_CTX_free(EVP_MD_CTX *ctx)
...
 {
-	image->md = EVP_sha384();
-	image->mdctx = EVP_MD_CTX_new();
+	assert(context);
+
+	if (context->context)
+		EVP_MD_CTX_free(context->context);
+
+	context->context = NULL;
+}
```
aborisovich, line 158: I would delete this `hash_free` function since `EVP_MD_CTX_free` does a null check inside its implementation,
Just call inline `EVP_MD_CTX_free(context->context)` everywhere, and then context->context = NULL;
They use it all the time like this in the [OpenSSL 1.1.1](https://github.com/openssl/openssl/tree/OpenSSL_1_1_1-stable) branch and newer 3.0+.
Also it does not make sense to keep 2 line function named "hash_free" that assigns a NULL to pointer freed, since
lexically this is not related to freeing the memory operation. Thus, I would just remove this function.

marcinszkudlinski (dismissed):

aborisovich (changes requested):
First look from me, I'll dig into the code tomorrow some more.

src/misc_utils.c
softwarecki: I just moved this function from another file. Git blamed Liam of adding it in this [commit](https://github.com/thesofproject/rimage/commit/b537522a6501f0f7249f26c24eb1a4d8679f4bf1). This commit moves all the code from tools repository, so I'm not sure who is the author. I will remove author entry.

src/adsp_config.c
```diff
@@ -18,6 +18,8 @@
 #include "toml.h"
 #include <stdbool.h>
 #include <stdint.h>
+#include <stdarg.h>
+#include <stdlib.h>
```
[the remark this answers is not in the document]
softwarecki: Apparently these files include `rimage.h` from which I removed the includes of openssl headers. Apparently the openssl headers included some standard libraries.

src/hash.c
pjdobrowolski: @aborisovich do you agree now?

aborisovich (approved):
It slowly starts to look like a program, good job ;-)
I assume you had tested it on both windows and linux yes?

marc-hb:
This does not compile, fix submitted in #144

marc-hb:
It's pretty easy to submit a temporary/draft/throw-away `sof/west.yml` test PR and make SOF CI run the FULL SOF test suites on test rimage changes BEFORE they're merged.

PR #143, opened 2023-03-06
mtl: add eq-iir and fir support

RanderWang:
Port from tgl-cavs.toml

fixes bug : https://github.com/thesofproject/sof/issues/7204 ([BUG] missing IIR EQ module config in rimage for MTL)

### DIFF
--- a/config/mtl.toml
+++ b/config/mtl.toml
@@ -57,7 +57,7 @@ name = "ADSPFW"
 load_offset = "0x40000"
 
 [module]
-count = 16
+count = 18
 	[[module.entry]]
 	name = "BRNGUP"
 	uuid = "2B79E4F3-4675-F649-89DF-3BC194A91AEB"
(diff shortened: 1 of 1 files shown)


marc-hb:
The size of these configuration files seems to be spiralling out of control.

They should probably have not been located in the rimage.git repo but in the sof.git repo instead: isn't the dependency on sof.git tighter than on rimage.git code?

kv2019i (approved):
We need this to solve a regression, but I agree with @marc-hb this is starting to be a bottleneck. Zooming out, it would seem we could just move the per-platform toml files to SOF repo. This would seem more practical --- rimage provides the infra, but content (what goes to a SOF image for a particular platform/product) is done on SOF side.

We also have the documentation part still WIP in https://github.com/thesofproject/sof-docs/pull/451/files

singalsu (approved):

lgirdwood (approved):

PR #144, opened 2023-03-06
rimage: fix build error

aiChaoSONG:
Recent changes introduce build error, fix the build error.

### DIFF
--- a/src/pkcs1_5.c
+++ b/src/pkcs1_5.c
@@ -312,7 +312,7 @@ static int rimage_verify(EVP_PKEY *privkey, enum manver ver,struct hash_context 
 	}
 
 	switch (ver) {
-	case V15	/* fallthrough */
+	case V15:	/* fallthrough */
 	case V18:
 		break;
 
@@ -337,8 +337,6 @@ static int rimage_verify(EVP_PKEY *privkey, enum manver ver,struct hash_context 
 		ERR_error_string(ERR_get_error(), err_buf);
 		fprintf(stderr, "error: verify %s\n", err_buf);
 	}
-
-	break;
 
 out:
 	EVP_PKEY_CTX_free(ctx);
--- a/src/css.c
+++ b/src/css.c
@@ -7,6 +7,7 @@
 
 #include <stdio.h>
 #include <errno.h>
+#include <time.h>
 #include <sys/time.h>
 #include <rimage/rimage.h>
 #include <rimage/css.h>


src/pkcs1_5.c
```diff
@@ -338,8 +338,6 @@ static int rimage_verify(EVP_PKEY *privkey, enum manver ver,struct hash_context
 		fprintf(stderr, "error: verify %s\n", err_buf);
 	}
 
-	break;
-
 out:
```
dbaluta, line 341: why do you remove the break? It doesn't seem to belong to this patch if you really want to remove it.

softwarecki (approved):

src/pkcs1_5.c
juimonen: the break is in the middle of a function outside switch statement... so probably some copy/paste error, probably better just to remove. I mean this break was introduced in the same patch @aiChaoSONG is trying to fix here.

lgirdwood (commented):
@marc-hb IIRC did we had a GH action to build rimage ?

kv2019i (approved):
Thanks @aiChaoSONG . Would be nice to add a reference to the commit that added the regression (d258ea23fc4d7). Obviously some basic build tests would be needed here. Now that we are using west for rimage, we'll find issues like this much later. And e.g. today, we'd have a critical update for rimage, but we can update rimage as the current head doesn't build. :(

src/pkcs1_5.c
lgirdwood: ok, so we are fixing a copy/paste error in the previous patch .

lgirdwood (approved):

marc-hb:
Actually, it's pretty easy to submit a temporary/draft/throw-away `sof/west.yml` test PR and make SOF CI run the FULL SOF test suites on test rimage changes BEFORE they're merged.

So maybe have more `rimage` CI is a bad idea after all, because it will _only_ test rimage compilation and make people feel like it's enough.

softwarecki:
I think that testing through sof CI will not solve the problem either. We have sections of code that depend on openssl version. The error appeared into the section for a different version of openssl than is used in rimage CI  I think that sof CI also uses only one openssl version. Rather, we should expand CI to use several openssl versions and perform image signatures and verification for different platforms.

marc-hb:
In general I agree you can never have too much validation and too much test coverage - except for limited time and resources.

In this case though we should enforce a single openssl version per product. Maybe even a single openssl version across all SOF.

Why? Simply because we have very limited validation time and resources.

@aiChaoSONG did you find this failure in an unsupported openssl version?

KokoSoft:
It looks like @aiChaoSONG is using openssl version above 3.0. If rimage pass CI in this PR https://github.com/thesofproject/sof/pull/7196 , it means that it using older openssl same as rimage CI.

lgirdwood:
We need to deprecate support for older ssl versions. Version 3.0 has been out for some time now so should be easily available on all developer OS configurations.

marc-hb:
tl;dr: openssl3 reached popular distros about 1 year ago but there is no stable Debian release with openssl 3 yet :-(
https://packages.debian.org/search?searchon=names&keywords=openssl

Ubuntu 22.04 was the first Ubuntu release with openssl3
https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=openssl&searchon=names
https://manpages.ubuntu.com/manpages/impish/man1/openssl.1ssl.html

Fedora apparently upgraded to openssl 3 in Fedora 36 which was released in May 2022
https://bodhi.fedoraproject.org/updates/?packages=openssl&page=2
Fedora 35 is not maintained any more.

kv2019i:
@softwarecki @marc-hb And tackling different openssl versions is a bit of separate issue. Our PR CI will only use one version at time and a minimum bar is that any submitted code works with _that_ version. If it doesn't, the code will break the CI for all new submitted code.

PR #148, opened 2023-03-10
Fix Zephyr IPC3 builds

lyakh:
Zephyr IPC3 builds use the same TOML files as XTOS builds, therefore those files should contain alias specifiers.

### DIFF
--- a/src/include/rimage/rimage.h
+++ b/src/include/rimage/rimage.h
@@ -138,6 +138,11 @@ struct fw_image_manifest_module {
 	struct sof_man_module *mod_man;
 };
 
+enum rtos_variant {
+	RTOS_XTOS,
+	RTOS_ZEPHYR,
+};
+
 /*
  * Audio DSP descriptor and operations.
  */
@@ -146,6 +151,7 @@ struct adsp {
 	struct mem_zone mem_zones[SOF_FW_BLK_TYPE_NUM];
 
 	uint32_t image_size;
+	enum rtos_variant image_rtos;
 	uint32_t dram_offset;
 
 	uint32_t alias_cached;
--- a/config/tgl-h.toml
+++ b/config/tgl-h.toml
@@ -3,6 +3,7 @@ version = [2, 5]
 [adsp]
 name = "tgl"
 image_size = "0x1F0000" # (30 + 1) bank * 64KB
+alias_mask = "0xE0000000"
 
 [[adsp.mem_zone]]
 type = "ROM"
@@ -16,6 +17,13 @@ size = "0x100000"
 type = "SRAM"
 base = "0xBE040000"
 size = "0x100000"
+
+[[adsp.mem_alias]]
+type = "uncached"
+base = "0x9E000000"
+[[adsp.mem_alias]]
+type = "cached"
+base = "0xBE000000"
 
 [cse]
 partition_name = "ADSP"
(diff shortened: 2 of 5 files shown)


lyakh:
if this fix is acceptable it should unlock https://github.com/zephyrproject-rtos/zephyr/pull/55554 and in general I think it wouldbe good for rimage to be aware of the RTOS type used to build the image

src/include/rimage/rimage.h
```diff
@@ -138,6 +138,11 @@ struct fw_image_manifest_module {
 	struct sof_man_module *mod_man;
 };
 
+enum rtos_variant {
+	RTOS_XTOS,
```
lgirdwood, line 142: Rather than call this Zephyr/xtos and be RTOS specific, this is really about how we handle cache memory mappings. Can we name this to refer to the different ways we handle memory.

lyakh: @lgirdwood well, at a low level that's about how linker scripts are written. Zephyr linker scripts use cached aliases for certain sections and uncached for others. Our XTOS linker scripts don't do that, all sections there have cached addresses. I thought it could be useful for rimage to know what kind of an image it's handling, not only for address mangling. But we can make it a memory type flag too, sure.
But in general, in principle it should be possible to always use the "Zephyr version" - to always convert addresses. Converting cached aliases to cached again shouldn't change them. I'm testing that version now too #149 + https://github.com/thesofproject/sof/pull/7226 - if that works, maybe we need the first patch in this PR

softwarecki:
I have mixed feelings about this pr. On the one hand, we get rid of the odd script from the build process and on the other, we further complicate the already unreadable rimage code. Personally, I would prefer rimage not to recognize the image type and not to manipulate addresses. If it really needs to do this, then maybe we can add a parameter to indicate that this should be done.

lgirdwood:
> I have mixed feelings about this pr. On the one hand, we get rid of the odd script from the build process and on the other, we further complicate the already unreadable rimage code. Personally, I would prefer rimage not to recognize the image type and not to manipulate addresses. If it really needs to do this, then maybe we can add a parameter to indicate that this should be done.

I assume you mean an enable/disable flag in the toml alongside any metadata ?

softwarecki:
I meant about the flag in the command line of the program, indicating zehyr image or the need to correct addresses. 

I read the description posted in the fix_elf_addrs.py file. Do I understand correctly that this is just a workaround for some rimage limitations? It is not about the firmware loading process? If so, I agree that we should remove this script and improve rimage. It seems to me that we should teach rimage to fully handle noncachable addresses instead of adding rtos recognition.

lyakh:
> I meant about the flag in the command line of the program, indicating zehyr image or the need to correct addresses.
> 
> I read the description in the fix_elf_addrs.py file. Do I understand correctly that this is just a workaround for some rimage limitations? It is not about the firmware loading process? If so, I agree that we should remove this script and improve rimage. It seems to me that we should teach rimage to fully handle noncachable addresses instead of adding rtos recognition.

@softwarecki in fact I don't think we need the commit, checking the RTOS type, that's why I made #149 and tested it with https://github.com/thesofproject/sof/pull/7226 so, now that #149 has been merged, let me close this one and then see, if we need any more address space intelligence in rimage.

PR #149, opened 2023-03-14
Revert "config:tgl/tglh: Do not set cached/uncached address aliases"

lyakh:
Now that rimage can distinguish between Zephyr and XTOS builds, we can use memory alias specifiers with both.

This reverts commit 1e0a85b44a19db37f9dd18906091d87abd8bd5e6.

### DIFF
--- a/config/tgl-h.toml
+++ b/config/tgl-h.toml
@@ -3,6 +3,7 @@ version = [2, 5]
 [adsp]
 name = "tgl"
 image_size = "0x1F0000" # (30 + 1) bank * 64KB
+alias_mask = "0xE0000000"
 
 [[adsp.mem_zone]]
 type = "ROM"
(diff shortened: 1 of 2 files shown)


lyakh:
It looks like this is enough now, as https://github.com/thesofproject/sof/pull/7226 confirms. I think previously this didn't work, which was the reason why I added https://github.com/thesofproject/rimage/pull/148/commits/054f8932cce774a1e18a4161429750be6d8f7df6 in #148 but apparently we've fixed something since then and now it isn't needed any more. If we do hit errors again, we can re-add it or something similar

lgirdwood (approved):
Do we need one for MTL ?

lgirdwood:
@kv2019i pls rview

kv2019i (approved):

lyakh:
> Do we need one for MTL ?

@lgirdwood mtl doesn't support IPC3 AFAIK, there aliases have been set a few weeks back

kv2019i:
Thanks for doing the SOF test run! Looks good, merging.

PR #151, opened 2023-03-30
Set of small changes to improve code readability and fix bugs (SP2)

softwarecki:
* Added header guard in elf.h file to allow safe include this header by multiple units.
* The man create modules function may return an error that was ignored. Added check of a return value.
* There was no dram_offset value in any configuration file. It also wasn't used anywhere in the code. So the code reading it from the tolm file has been removed.
* Added functions to print enum and flags value. Added name_val structure representing name/value tuple. Added function
print_enum which displays the name of the enum corresponding to a given value. Added print_flags function which display names of flags constituting a given value.
* Create new memory_config structure. Moved mem_zone field from adsp structure to the new memory_config structure
* Moved memory alias configuration to a new struct. A new memory_alias structure containing memory address alias configurations has been created. It was placed in the memory_config structure.
* Rename struct module to manifest_module. Changing the name of the structure type is a prelude to separating information about the elf file from the information used during creation of a manifest.
* Renamed fw_size field in manifest_module structure to output_size. New name is a more clear. Added description for this field.
* A is_bootloader field has been added to determine whether a module is a bootloader. This is clearer than checking the module index and the number of modules in different places in the program.
* Add suport for loadable modules. A new -l parameter has been added to the program's command line. It instructs the program to not treat the first module as a bootloader and not to skip the bss section for it.
* Moved MAN_MAX_SIZE_V1_8 define to the public manifest header. Currently, sof incorrectly includes a private manifest.h header belonging to the manifest.c module. Moved the MAN_MAX_SIZE_V1_8 definition used by sof, to the public manifest header to allow sof to switch to using the correct file.

### DIFF
--- a/src/adsp_config.c
+++ b/src/adsp_config.c
@@ -66,7 +66,6 @@ static void dump_adsp(const struct adsp *adsp)
 	DUMP("\nadsp");
 	DUMP_KEY("name", "'%s'", adsp->name);
 	DUMP_KEY("image_size", "0x%x", adsp->image_size);
-	DUMP_KEY("dram_offset", "0x%x", adsp->dram_offset);
 	DUMP_KEY("exec_boot_ldr", "%d", adsp->exec_boot_ldr);
 	for (i = 0; i < ARRAY_SIZE(adsp->mem_zones); ++i) {
 		DUMP_KEY("mem_zone.idx", "%d", i);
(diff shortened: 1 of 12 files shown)


softwarecki:
SOF CI PR https://github.com/thesofproject/sof/pull/7350

lgirdwood (approved):
Good stuff @softwarecki !

PR #173, opened 2023-08-24
config: lnl+mtl: fix length of ADSP.man CSE manifest

kv2019i:
The ADSP.man contents is different for ACE_V1_5 layout compared to CAVS_V2_5 layout. so the ADSP.man length should be adjusted accordingly.

With this fix, sof_ri_info.py can be used to analyze LNL and MTL firmware images created with rimage.

Closes: https://github.com/thesofproject/sof/issues/8073

### DIFF
--- a/config/lnl.toml
+++ b/config/lnl.toml
@@ -30,7 +30,7 @@ partition_name = "ADSP"
 [[cse.entry]]
 name = "ADSP.man"
 offset = "0x5c"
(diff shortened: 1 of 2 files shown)


kv2019i:
Tested with https://github.com/thesofproject/sof/pull/8090 - tests pass.

abonislawski (approved):

PR #176, opened 2023-09-04
[RFC] rimage: Add support for multiple toml file

aiChaoSONG:
This patch adds support for multiple toml
configuration file.

### DIFF
--- a/src/adsp_config.c
+++ b/src/adsp_config.c
@@ -25,6 +25,7 @@
 #include <stdio.h>
 #include <errno.h>
 #include <ctype.h>
+#include <sys/stat.h>
 
 /* map memory zone string name to enum value */
 static enum snd_sof_fw_blk_type zone_name_to_idx(const char *name)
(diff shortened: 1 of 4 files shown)


aiChaoSONG:
The toml lib we use can not support toml merging, we need to merge multiple files together before parsing. This RFC used a function `fmemopen`, I am not sure if MSVC toolchain support it. I could also use temp file for merging if  `fmemopen` is not supported.

aiChaoSONG:
@kv2019i  @lgirdwood  refering to https://github.com/thesofproject/sof/issues/7270 and this RFC, could you please comment?

src/adsp_config.c
```diff
@@ -2337,18 +2338,69 @@ static int adsp_parse_config_fd(FILE *fd, struct image *image)
...
+		if (read != file_size[i]) {
+			fprintf(stderr, "error: failed to read file %s\n", files->file[i]);
+			return NULL;
+		}
+		fclose(file);
+
+		buf[file_size[i]] = '\n';
+		buf += file_size[i] + 1;
+	}
+	return fmemopen(files->buffer, buf_size, mode);
```
kv2019i, line 2384: Ok this is quite clever approach. "fmemopen()" will potentially hit some compatibility issues, so needs a smoketest on Windows at minimum.

I was actually thinking of something much more low-tech. I.e. "cat a.toml b.toml >$BUILD/rimage.toml" in make rules, and pass generated "rimage.toml" to rimage. But making this in a portable way is not zero effort either, so I guess depends in the end what is the lowest effort way to make this so that it works across platforms.

aiChaoSONG: A new proposal, see #179, this probably won't work on windows, let's close

marc-hb:
> I was actually thinking of something much more low-tech. I.e. "cat a.toml b.toml >$BUILD/rimage.toml" in make rules, and pass generated "rimage.toml" to rimage. But making this in a portable way is not zero effort either, 

Windows has `mktemp` too so it should really not be that hard to do in `rimage`.

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/mktemp-s-wmktemp-s?view=msvc-170

PR #178, opened 2023-09-07
mtl: remove pin description from extended module config

yongzhi1:
Extended module config such as pin description is not used by driver, let's remove them, lest anyone waste time on the don't care values when adding a new module.

### DIFF
--- a/config/mtl.toml
+++ b/config/mtl.toml
@@ -87,13 +87,6 @@ count = 22
 	load_type = "0"
 	module_type = "1"
 	auto_start = "0"
-	sched_caps = [1, 0x00008000]
-
-	# pin = [dir, type, sample rate, size, container, channel-cfg]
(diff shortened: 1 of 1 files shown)


lgirdwood (commented):
@mwasko @marcinszkudlinski will this be needed in the future ?

marcinszkudlinski:
We need at least number of pins for each module.
I would not remove it, 

now formats and pins are verified when pipeline starts (module_prepare), which may be a little too late. Bind IPC should not be accepted if connecting not compatible modules or not existing pins - based on info from the manifest.

So yes, those may become useful really soon.

yongzhi1:
Thanks for the info, I somehow got the impression that Extended Manifest ("$AE1" ) is not used by FW.

lgirdwood:
@yongzhi1 I will close this now. It can be re-opende later if direction changes.

PR #179, opened 2023-09-13
[RFC] rimage: Add support for multiple toml file

aiChaoSONG:
add support for multiple toml file for rimage, prepare for split toml file
to platform related toml and algorithm related toml.

The previous RFC #176 won't work for windows system, because it uses a linux specific function,
This RFC should work both for windows and linux.

### DIFF
--- a/src/adsp_config.c
+++ b/src/adsp_config.c
@@ -2337,15 +2337,50 @@ error:
 	return ret;
 }
 
+static void copy_file(FILE *src, FILE *dst)
+{
+	char ch;
+	while ((ch = fgetc(src)) != EOF)
+		fputc(ch, dst);
+}
+
+static FILE* adsp_conf_files_open(const struct adsp_conf_files *files, const char *mode)
+{
+	FILE *file, *out_file;
+	int i;
+
+	out_file = fopen(files->out_file, "w");
+	if (!out_file) {
+		file_error("unable to open file for writing", files->out_file);
+		return NULL;
+	}
+
+	/* Merge multiple input toml files into one */
+	for (i = 0; i < files->file_count; i++) {
+		fprintf(out_file, "## File path: %s\n", files->file[i]);
+		file = fopen(files->file[i], "r");
+		if (!file) {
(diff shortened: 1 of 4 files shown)


src/rimage.c
```diff
@@ -122,8 +122,8 @@ int main(int argc, char *argv[])
 		return -EINVAL;
 	}
 
-	/* make sure we have an outfile if not verifying */
-	if ((!image.out_file && !image.verify_file)) {
+	/* We are either signing or verifying an image, not both or neither */
+	if (!(!!image.out_file ^ !!image.verify_file)) {
```
btian1, line 131: possible to figure out a more easier expression for better understanding?

src/adsp_config.c
```diff
@@ -2337,15 +2337,50 @@ static int adsp_parse_config_fd(FILE *fd, struct image *image)
...
+	/* Merge multiple input toml files into one */
+	for (i = 0; i < files->file_count; i++) {
+		fprintf(out_file, "## File path: %s\n", files->file[i]);
+		file = fopen(files->file[i], "r");
+		if (!file) {
+			file_error("unable to open file for reading", files->file[i]);
+			fclose(out_file);
+			return NULL;
+		}
+		copy_file(file, out_file);
```
btian1, line 2367: use file pointer as parameter is not always proposed, directly read and write is also very simple.
btw, how about fread/fwrite? since toml file is not big.

src/rimage.c
lgirdwood: " not both or neither " would this not imply to fail if ```!out_file == !verify_file``` ? Pls check my logic here though.

lgirdwood (commented):
@aiChaoSONG btw, this can be better done with the GNU pre-processor (like we already pre-process linker scripts).
i.e. mtl-sku.toml
```
// all platform memoery addresses
#include <platform.toml>
// all Intel signing config
#include <intel.toml>
// all modules 
#include <volume.toml>
#include <copier.toml>
and so on
```

PR #184, opened 2023-10-13
mtl: remove smart amp config in basefw

RanderWang:
Smart amp will be a loadable module.

It is only for mtl-006-stable branch. Since rimage was integrated into sof, this change will not affect main branch.

### DIFF
--- a/config/mtl.toml
+++ b/config/mtl.toml
@@ -57,7 +57,7 @@ name = "ADSPFW"
 load_offset = "0x40000"
 
 [module]
-count = 25
+count = 24
 	[[module.entry]]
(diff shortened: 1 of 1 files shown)


mengdonglin (approved):

pjdobrowolski (approved):

kv2019i:
This should have gone to https://github.com/thesofproject/rimage/tree/mtl-006-drop-stable branch as a PR. Sorry, bad guidelines from me earlier. This repo is no longer used by SOF mainline, so this PR did not cause any harm either, but still to use this in mtl-006, it needs to be submitted to correct branch.

PR #191, opened 2023-10-30
loadable module: define a new bit in fw_image_flags

RanderWang:
The bit 4 is set to force driver reload library on d3 exit.

It is a prototype for validation with mtl-006 branch so I use hard core.  We need to add a new feature that toml can decode macro definition like
fw_image_flag = BASE_FW | RELOAD_LIBRARY

Kernel PR https://github.com/thesofproject/linux/pull/4669 depends on this rimage update.

### DIFF
--- a/config/mtl.toml
+++ b/config/mtl.toml
@@ -55,6 +55,7 @@ base_offset = "0x2000"
 [fw_desc.header]
 name = "ADSPFW"
 load_offset = "0x40000"
+fw_image_flags = "0x12" # BASE_FW (bit 1) + CONTEXT_IS_NOT_SAVED (bit 4)
 
 [module]
 count = 18
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -1463,7 +1463,6 @@ int man_write_fw_ace_v1_5(struct image *image)
 	m->desc.header.build_version = image->fw_ver_build;
 
 	m->desc.header.feature_mask = 0x2; // -> should be feature mask - to fix
-	m->desc.header.fw_image_flags = 0x2; // -> should be feature mask - to fix
 	m->desc.header.fw_compat = 0x100000; // -> PUT PROPER STRUCT
 
 	/* create each module */


src/manifest.c
```diff
@@ -1463,7 +1463,6 @@ int man_write_fw_ace_v1_5(struct image *image)
 	m->desc.header.build_version = image->fw_ver_build;
 
 	m->desc.header.feature_mask = 0x2; // -> should be feature mask - to fix
-	m->desc.header.fw_image_flags = 0x2; // -> should be feature mask - to fix
```
btian1, line 1466: previous implementation have issues? if not, change 0x2 --> 0x12 will make changes more less.

RanderWang: it is fixed and it should be built from .toml file

config/mtl.toml
```diff
@@ -55,6 +55,7 @@ base_offset = "0x2000"
 [fw_desc.header]
 name = "ADSPFW"
 load_offset = "0x40000"
+fw_image_flags = "0x12"
```
ujfalusi, line 58: Can you add a comment that `0x12` is `BASE_FW`+`CONTEXT_IS_NOT_SAVED` or something like this?

RanderWang: thanks, updated

RanderWang:
add more comments

lgirdwood:
+ @mmaka1 for review

mmaka1:
The ability to survive D3 is the property of FW, not the FW image binary (a file). It is known at the FW build time, therefore adding it to the binaries making this information instantly available to the driver looks like a good idea. But then it should be added consistently to all the binaries, otherwise it might be interpreted like a different behavior of individual binaries (base_fw/libs) while this defines a behavior of the FW related to all loaded binaries. So, it requires extra maintenance and looking from this perspective, adding this property to the IPC4 FW_CONFIG parameter of the Base FW looks like a simpler solution to maintain (adding more overhead at the run-time though) unless we agree to develop an asymmetric Extended Manifest containing properties valid for the Base FW only.

marc-hb:
Sorry for adding another twist to an already complex topic but... why is this submitted directly to `mtl-006-drop-stable`? Shouldn't complex and long term design questions like this be debated and more importantly _tested_ on the main development branch first before going to a maintenance branch?

RanderWang:
no need

